PR #5, opened 2018-12-21
Working builds

amckague:
Using https://developer.github.com/v3/repos/statuses/ to wrap the build running in bamboo + commit vendor


ash2k:
The commits here don't fully make sense to me =) I'll try to do vendor + everything else locally to see the picture better.

PR #6, opened 2018-12-24
Autowiring v2, step 1

ash2k:


pkg/execution/plugins/iamrole/iam_role_manager.go
```diff
@@ -95,7 +96,7 @@ var defaultEC2ComputeAssumeRoleStatement = IamAssumeRoleStatement{
 	Action: "sts:AssumeRole",
 }
 
-func defaultJSON(serviceID string) *IamPolicy {
+func defaultJSON(serviceID voyager.ServiceName) *IamPolicy {
```
nilebox: rename variable to `serviceName`?

pkg/execution/plugins/iamrole/types.go
```diff
@@ -14,7 +15,7 @@ const (
 )
 
 type Spec struct {
-	ServiceID             string                 `json:"serviceId,omitempty"`
+	ServiceID             voyager.ServiceName    `json:"serviceId,omitempty"`
```
nilebox: so is it a name or an ID? in JSON it's still `serviceId`...

pkg/orchestration/wiring/entangler.go
```diff
@@ -99,13 +100,12 @@ func (en *Entangler) Entangle(state *orch_v1.State, context *orchestration.Entan
 		return nil, false, err
 	}
 
-	tags := make(map[voyager.Tag]string, len(serviceProperties.UserTags))
+	tags := make(map[voyager.Tag]string, len(serviceProperties.UserTags)+6)
```
nilebox: why 6? is this optimization worth it? just create an empty map instead?
alternatively, create a separate "static" map with all 6 known elements, i.e.
```go
platformTags := map[voyager.Tag]string{
  en.TagNames.ServiceNameTag: string(context.ServiceName),
  en.TagNames.BusinessUnitTag: serviceProperties.BusinessUnit,
  ...
}
tags := make(map[voyager.Tag]string, len(serviceProperties.UserTags)+len(platformTags))
...
```

pkg/orchestration/wiring/wiringplugin/types.go
```diff
@@ -11,34 +11,90 @@ import (
...
+	Contract ResourceContract
+	// DEPRECATED: use Contract
 	SmithResources []smith_v1.Resource
-	Attributes     map[string]interface{}
+	// Attributes are attributes attached to the edge between resources.
+	Attributes map[string]interface{}
+}
+
+// Dependant represents a resource that depends on the resource that is currently being processed.
+type Dependant struct {
```
nilebox: DependantResource?

pkg/orchestration/wiring/wiringplugin/types.go
```diff
@@ -11,34 +11,90 @@ import (
...
 
 type WiringResult struct {
+	Contract  ResourceContract
 	Resources []WiredSmithResource
 }
 
 type WiredSmithResource struct {
 	SmithResource smith_v1.Resource
-	Exposed       bool
+	// DEPRECATED: use shapes instead
```
nilebox: there is no shapes (yet?) to use instead of Exposed

pkg/execution/plugins/iamrole/types.go
ash2k: It needs to be name, not ID

pkg/orchestration/wiring/wiringplugin/types.go
ash2k: They are in the `ResourceContract` which is a field in the `WiringResult`

nilebox: it's not clear from this comment

pkg/orchestration/wiring/entangler.go
ash2k: +6 because there are 6 more elements that are added to the map below.

pkg/execution/plugins/iamrole/types.go
nilebox: rename the field and vars for now and leave a TODO for renaming in JSON? or deprecate this field and add a `ServiceName` as a new field, and eventually migrate to it.

pkg/orchestration/wiring/entangler.go
nilebox: so we will have to update this magic number every time we add/remove tag. see my suggestion to create a separate `platformTags` map and allocate `tags` size based on it (and copy platform tags from it).

pkg/execution/plugins/iamrole/types.go
ash2k: I think we need to change these plugins to be example code rather than real code. This refactoring needs to be done in the internal codebase.

pkg/orchestration/wiring/entangler.go
ash2k: I don't want to allocate 2 temporary maps just to merge them into a third one. I'll just remove the number if you don't like it so much =)

nilebox: up to you :)

ash2k:
This is ready for review

PR #13, opened 2018-12-30
Create common secret during synchronization

Burr:
* Create empty secret during namespace creation
* Operation is create-only, should not update


pkg/synchronization/controller.go
```diff
@@ -534,6 +539,38 @@ func (c *Controller) createOrUpdateSecret(logger *zap.Logger, secretSpec *core_v
 	return true, secret, err
 }
 
+func (c *Controller) ensureCommonSecretExists(logger *zap.Logger, ns *core_v1.Namespace) (bool /* retriable */, error) {
+	_, err := c.MainClient.CoreV1().Secrets(ns.Name).Get(apik8scompute.CommonSecretName, meta_v1.GetOptions{})
+
+	// nothing to do, it exists; nothing to update
+	if err == nil {
+		logger.Sugar().Infof("Secret named %q already exists", apik8scompute.CommonSecretName)
```
nilebox: To think about in future: the secret may be outdated?

pkg/synchronization/controller.go
```diff
@@ -534,6 +539,38 @@ func (c *Controller) createOrUpdateSecret(logger *zap.Logger, secretSpec *core_v
...
+			},
+			TypeMeta: meta_v1.TypeMeta{
+				Kind:       k8s.SecretKind,
+				APIVersion: core_v1.SchemeGroupVersion.String(),
+			},
+			Type: core_v1.SecretTypeOpaque,
+		}
+
+		_, err = c.MainClient.CoreV1().Secrets(ns.Name).Create(secret)
+		return true, err
```
nilebox: I know it won't have any effect, but I would write
```go
if err != nil {
  return true, err
}
return false, nil
```

pkg/synchronization/controller.go
```diff
@@ -534,6 +539,38 @@ func (c *Controller) createOrUpdateSecret(logger *zap.Logger, secretSpec *core_v
...
+	}
+
+	// unknown error
+	return false, err
```
nilebox: Could be retriable? e.g. in case of request timeout or some other temporary network glitch

nilebox (dismissed):

pkg/synchronization/controller.go
Burr: In general since it has no contents at creation time, it can't really be outdated. Or is the worry that it somehow hangs around if they delete/recreate everything? If the namespace gets deleted, the secret should be deleted, too.

nilebox: sorry, I missed that we create an empty secret, all good.

pkg/synchronization/controller.go
```diff
@@ -534,6 +539,38 @@ func (c *Controller) createOrUpdateSecret(logger *zap.Logger, secretSpec *core_v
 	return true, secret, err
 }
 
+func (c *Controller) ensureCommonSecretExists(logger *zap.Logger, ns *core_v1.Namespace) (bool /* retriable */, error) {
+	_, err := c.MainClient.CoreV1().Secrets(ns.Name).Get(apik8scompute.CommonSecretName, meta_v1.GetOptions{})
```
ash2k: Use package level constant you defined =) Here are in other places.

pkg/synchronization/controller.go
ash2k: This should not log at all, or do so on debug to avoid spamming too much.

pkg/synchronization/controller.go
```diff
@@ -534,6 +539,38 @@ func (c *Controller) createOrUpdateSecret(logger *zap.Logger, secretSpec *core_v
...
+
+	// nothing to do, it exists; nothing to update
+	if err == nil {
```
ash2k: I think it might be better to use `switch` here to show that there are a few distinct cases to handle. WDYT?

pkg/synchronization/controller.go
Burr: Updated.

pkg/synchronization/controller.go
Burr: Updated.

pkg/synchronization/controller.go
```diff
@@ -534,6 +539,41 @@ func (c *Controller) createOrUpdateSecret(logger *zap.Logger, secretSpec *core_v
 	return true, secret, err
 }
 
+func (c *Controller) ensureCommonSecretExists(logger *zap.Logger, ns *core_v1.Namespace) (bool /* retriable */, error) {
```
ash2k: Probably doesn't need to whole `ns` object, but I guess you are just following the style used in other functions. np

pkg/synchronization/controller.go
Burr: I agree that might be slightly cleaner. Is there a good way to structure that with `api_errors.IsNotFound` though?

ash2k: https://github.com/golang/go/wiki/Switch#missing-expression

pkg/synchronization/controller.go
Burr: Updated to Debug.

pkg/synchronization/controller.go
Burr: 🤦‍♂️ Updated

pkg/synchronization/controller.go
Burr: Updated

ash2k (approved):

jokeyrhyme (approved):

PR #19, opened 2019-01-01
Use ServiceName type in more places for safety

ash2k:


pkg/servicecentral/types.go
```diff
@@ -1,6 +1,13 @@
 package servicecentral
 
-import "time"
+import (
+	"time"
+)
+
+// ServiceName is the name of a Service.
+// Not voyager.ServiceName because not all Services in Service Central are Voyager services.
```
nilebox: I understand the argument, but still think that it's fine to just use `voyager.ServiceName`:
1. `voyager.ServiceName` doesn't necessarily guarantee that the service is a Voyager service - it's just a type that says "this is a service name, not just a random string", and every Voyager service name is a Service Central service name, so semantically they are the same
2. In this SC client we have fiiltering to fetch only voyager services (via `platform` tag), so in this client every service is actually a Voyager service.

ash2k: Yeah, that is actually how I did it initially. But then I thought that if we'd move that client out of tree then we will likely have a different type there - because it shouldn't depend on the voyager repo.

PR #26, opened 2019-01-03
Re-generated (re-formatted) expected entangler test bundles

kopper:


ash2k:
We probably need to have a unit test that validates the fixtures are formatted properly.

PR #27, opened 2019-01-03
Added e2e tests state to entangler tests

kopper:


ash2k:
Merged into wrong branch, but too later. Should be `master`. I'll reopen as another PR.

PR #29, opened 2019-01-03
Move CRDs into separate packages

ash2k:
With this change only `cmd/crd` and creator depend on apiextensions-server dependency.


nilebox:
> creator depend on apiextensions-server dependency.

@ash2k Why does creator depend on apiextensions-apiserver sorry? IIRC creator doesn't use any CRDs...

ash2k:
Sorry, not creator. It's the admission hook in replication.

PR #30, opened 2019-01-03
Made DynamoDB to export ResourceContract and migrated all consumers t…

kopper:
…o use it


pkg/orchestration/wiring/aws/plugin.go
```diff
@@ -37,13 +37,14 @@ func Resource(resourceType voyager.ResourceType,
 	clusterServicePlanExternalID servicecatalog.PlanExternalID,
 	generateServiceEnvironment ServiceEnvironmentGenerator,
 	envResourcePrefix oap.EnvVarPrefix,
+	outputResourceContract bool,
```
ash2k: See my branch https://github.com/atlassian/voyager/compare/svccat-shapes

I think it is not do it or don't do it, it's which shape to output. Same here and in svc cat entangler. This is because some resources coming via OAP are not IAM accessible (not the case right now, but might be in the future).

Why don't you want to convert all of them at once? Why this flag per resource?

pkg/orchestration/wiring/ec2compute/common/common.go
```diff
@@ -139,11 +141,15 @@ func WireUp(microServiceNameInSpec, ec2ComputePlanName string, stateResource *or
 
 	for _, dependency := range dependencies {
 		if !dependency.Contract.IsEmpty() {
```
ash2k: I think this should work differently. It should look for shapes it is interested in and if not found fall back to current behavior. Then once everything exposes a contract, we remove the fallback and return an error instead.

pkg/orchestration/wiring/ec2compute/common/common.go
```diff
@@ -139,11 +141,15 @@ func WireUp(microServiceNameInSpec, ec2ComputePlanName string, stateResource *or
...
-			if dependency.Type != ups.ResourceType {
+			if dependency.Type != ups.ResourceType && dependency.Type != aws.DynamoDB {
```
ash2k: The idea of shapes is that you don't need to look at resource types.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -132,6 +135,21 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
 
 	for _, dep := range context.Dependencies {
 		found := false
+		if !dep.Contract.IsEmpty() {
```
ash2k: See comments for ec2 func

pkg/orchestration/wiring/wiringutil/knownshapes/bindable_environment_variables.go
```diff
@@ -1,19 +1,41 @@
...
+import (
+	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
+	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin"
+)
 
 const (
 	BindableEnvironmentVariablesShape wiringplugin.ShapeName = "voyager.atl-paas.net/BindableEnvironmentVariables"
 )
 
+type ServiceInstanceReference interface {
```
ash2k: This is not needed =) You only need this because you've made the type private. It shouldn't be for this very reason - you cannot use it from another package. But that is the point - to be able to import the type and use it for type safety. Assuming you are writing Go, of course.

ash2k: Also, this is Go anti pattern. Consuming side should be defining interfaces typically, not the producing side. It's not Java =)

pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
```diff
@@ -109,14 +110,18 @@ func (e *SvcCatEntangler) constructServiceInstance(resource *orch_v1.StateResour
 	}, nil
 }
 
-func (e *SvcCatEntangler) constructResourceContract(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) (wiringplugin.ResourceContract, error) {
+func (e *SvcCatEntangler) constructResourceContract(resource *orch_v1.StateResource, smithResource smith_v1.Resource, context *wiringplugin.WiringContext) (wiringplugin.ResourceContract, error) {
```
ash2k: See my branch https://github.com/atlassian/voyager/compare/svccat-shapes#diff-acbf2382321568395dd65543b8ca0b58R124

You don't know if both shapes should be produced.

pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
```diff
@@ -109,14 +110,18 @@ func (e *SvcCatEntangler) constructServiceInstance(resource *orch_v1.StateResour
...
 	// TODO(kopper): Actually implement.
 	return wiringplugin.ResourceContract{
-		Data: []wiringplugin.DataItem{
-			{
-				Name: "fakeNotEmptyContract",
-			},
+		Shapes: []wiringplugin.Shape{
+			knownshapes.NewBindableEnvironmentVariables(smithResource.Name),
+			knownshapes.NewBindableIamAccessible(smithResource.Name, "IamPolicySnippet"),
```
ash2k: path should be a parameter to `SvcCatEntangler` as it may be different.

pkg/orchestration/wiring/wiringplugin/types.go
```diff
@@ -85,6 +86,16 @@ type ResourceContract struct {
...
 }
 
+func (c *ResourceContract) GetShape(shapeName ShapeName) (Shape, error) {
+	for _, shape := range c.Shapes {
+		if shape.Name() == shapeName {
+			return shape, nil
+		}
+	}
+
+	return nil, errors.Errorf("Shape %s not found in the contract", shapeName)
```
ash2k: It's not an error for a shape to not be there, should be just `bool`.

pkg/orchestration/wiring/aws/plugin.go
kopper: This is WIP. Only UPS and DynamoDB are currently supported. Will add more resources later.

I decided to do incremental development and I stick to it. Sure you could do everything in one PR. Given that today I learned that this whole refactoring is much smaller change than I initially thought you might be right that this approach is overkill, but 🤷‍♂️ 

I'll see how many changes are required to support more resources. If not many I might support all remaining resources at once.

pkg/orchestration/wiring/ec2compute/common/common.go
kopper: I use outputResourceContract to mark resources that are migrated to contracts already, but I am not consistent and use if resourceType = X in the same "if". I should remove "if dependency.Type" check. It is unnecessary.

pkg/orchestration/wiring/ec2compute/common/common.go
kopper: Agreed. This is unnecessary check. I'll remove it.

pkg/orchestration/wiring/wiringutil/knownshapes/bindable_environment_variables.go
kopper: Good point for the latter comment. Bad habits 🤣 

As for why we made type private - because ShapeName is constant, we didn't want to allow to instantiate struct with arbitrary shape name. Alternatively we discussed changing Contract to contain a map of name -> Shape instead of array of Shapes.

I like constructor approach because ATM we use only some small parts of all available properties, so it makes it easier to instantiate it. But I have no strong opinion about this particular thing, I'm happy to change if we decide to use map.

pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
kopper: For now we don't need NewBindableIamAccessible shape yet, I'll remove it. I'll add it back when working on refactoring of IAM role plugin.

pkg/orchestration/wiring/wiringplugin/types.go
kopper: Error is defensive version of panic (I don't want to use panic because it will affect all states not only the problematic ones). I assume that if this method is called the shape is 100% there and there is exactly one, that is why I didn't name the method "find". I'll add comment.

pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
ychen-atlassian: Actually I don't think `SvcCatEntangler` should have any idea about IamRoles.

Consider the case where we have a new set of providers that produce some sort of `KerberosyTokenThings`. Eventually our code will look like this:

```
	IsEnvVarBindable bool
	IsIAMAccessible  bool
	IsKerberosAccesssible bool
	IsTokenProducing  bool
	IsAnotherShape bool
	IsReallyReallyCool  bool
```

I think @kopper current implementation makes sense - a `SvcCatEntangler` should *always* produce a `BindableEnvVar`. Actually, when we refactor this more, a boolean will not be sufficient - we'll need to force the consumers of `SvcCatEntangler` to provide us the actual functions that can generate (or know about) the data:

```
GetEnvVarData  func (ServiceInstance) BindableEnvVarData
GetBinaryData  func (ServiceInstance) BindableBinaryData
GetSecretData  func (ServiceInstance) BindableSecretData
```

ychen-atlassian: Egh replied to the wrong comment

pkg/orchestration/wiring/aws/plugin.go
ychen-atlassian: Also just giving some reasoning on why we shouldn't use IsIamAccessible

Consider the case where we have a new set of providers that produce some sort of `KerberosyTokenThings`. Eventually our code will look like this:

```
	IsEnvVarBindable bool
	IsIAMAccessible  bool
	IsKerberosAccesssible bool
	IsTokenProducing  bool
	IsAnotherShape bool
	IsReallyReallyCool  bool
```

I think @kopper current implementation makes sense - a `SvcCatEntangler` should *always* produce a `BindableEnvVar`. Actually, when we refactor this more, a boolean will not be sufficient - we'll need to force the consumers of `SvcCatEntangler` to provide us the actual functions that can generate (or know about) the data:

```
GetEnvVarData  func (ServiceInstance) BindableEnvVarData
GetBinaryData  func (ServiceInstance) BindableBinaryData
GetSecretData  func (ServiceInstance) BindableSecretData
AdditionalShapes func (ServiceInstance) Shapes[]
```

pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
ychen-atlassian: See comment about not providing IamAccessible - I don't think this util code should be aware of this shape at all.

pkg/orchestration/wiring/aws/plugin.go
kopper: Note this code is only for (some) aws resources, so the number of bool flags will be limited to AWS-specific stuff.

On consumer side (EC2Compute, KubeCompute) I check if contract contains specific shape (in this case envvarbindable).

ychen-atlassian: I think all our OSB providers currently use it at the moment. A provider (UPS for example) doesn't seem to need to know about AWS to provide the flag

pkg/orchestration/wiring/wiringutil/knownshapes/bindable_iam_accessible.go
```diff
@@ -10,13 +13,34 @@ const (
...
+func NewBindableIamAccessible(resourceName smith_v1.ResourceName, IAMPolicySnippetPath string) *BindableIamAccessible {
+	return &BindableIamAccessible{
+		ShapeMeta: wiringplugin.ShapeMeta{
+			ShapeName: BindableIamAccessibleShape,
+		},
+		Data: BindableIamAccessibleData{
+			BindableShapeStruct: wiringplugin.BindableShapeStruct{
+				ServiceInstanceName: wiringplugin.ProtoReference{
+					Resource: resourceName,
+					Path:     "metadata.name",
```
ash2k: Set `Example` please.

pkg/orchestration/wiring/wiringutil/knownshapes/bindable_environment_variables.go
```diff
@@ -9,11 +12,31 @@ const (
...
+	return &BindableEnvironmentVariables{
+		ShapeMeta: wiringplugin.ShapeMeta{
+			ShapeName: BindableEnvironmentVariablesShape,
+		},
+		Data: BindableEnvironmentVariablesData{
+			BindableShapeStruct: wiringplugin.BindableShapeStruct{
+				ServiceInstanceName: wiringplugin.ProtoReference{
+					Resource: resourceName,
+					Path:     "metadata.name",
+				}},
```
ash2k: Set `Example` please.

pkg/orchestration/wiring/wiringutil/service_binding.go
```diff
@@ -29,6 +30,20 @@ func ConsumerProducerServiceBinding(consumer, producer voyager.ResourceName, pro
...
 }
 
+// TODO(kopper): Remove V2 suffix
+func ConsumerProducerServiceBindingV2(consumer, producer voyager.ResourceName, resourceReference wiringplugin.ProtoReference,
+	exposed bool) wiringplugin.WiredSmithResource {
+
+	serviceInstanceRef := smith_v1.Reference{
+		Name:     ReferenceName(resourceReference.Resource, "metadata", "name"),
+		Resource: resourceReference.Resource,
+		Path:     resourceReference.Path,
```
ash2k: Must copy example and modifier fields too!

ash2k: I think we need to have a method on proto reference that takes missing pieces and produces a full smith reference. This will ensure we'll not forget some fields and that if we add new fields we'll get compilation errors in all places where we convert proto to a full ref.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -131,23 +132,38 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
...
+			resourceReference := bindableShape.(*knownshapes.BindableEnvironmentVariables).Data.ServiceInstanceName
+			// We don't want anything that depends on compute to see our bindings - exposed: false
+			binding := wiringutil.ConsumerProducerServiceBindingV2(resource.Name, dep.Name, resourceReference, false)
+			smithResources = append(smithResources, binding)
+			bindingResources = append(bindingResources, binding)
+			continue
+		}
+
+		// DEPRECATED: binding using raw SmithResources, use resource contracts instead
+		if len(dep.SmithResources) > 0 {
```
ash2k: This `if` is not needed

pkg/util/sets/BUILD.bazel
```diff
@@ -6,18 +6,13 @@ go_library(
         "clusterLocation.go",
         "doc.go",
         "empty.go",
-        "float32.go",
```
ash2k: This generator is annoying.

pkg/orchestration/wiring/wiringutil/service_binding.go
ash2k:
```go

// ToReference should be used to augment ProtoReference with missing information to
// get a full Reference.
func (r *ProtoReference) ToReference(name smith_v1.ReferenceName) *smith_v1.Reference {
	return &smith_v1.Reference{
		Name:     name,
		Resource: r.Resource,
		Path:     r.Path,
		Example:  r.Example,
		Modifier: r.Modifier,
	}
}
```
Add this method please.

pkg/orchestration/wiring/wiringutil/knownshapes/bindable_iam_accessible.go
kopper: Set it to what? I can't find any example of using this field in our codebase.

pkg/orchestration/wiring/wiringutil/service_binding.go
kopper: done

pkg/orchestration/wiring/k8scompute/plugin.go
kopper: removed

PR #33, opened 2019-01-04
Remove BAZEL_OPTIONS from query

ychen-atlassian:


ash2k:
What is the reason for this change? Is something not working?

ychen-atlassian:
Thought this was the problem with the builds, just a spurious error message. Fixing it in a different way

PR #35, opened 2019-01-05
Build race binaries instead of pure ones by default

ash2k:
Tests are built in race mode. If binaries are built in pure mode then all libraries are built twice.

Look at number of actions and elapsed time. Both runs were executed after bazel clean.

Before:
```console
bazel test \
                --test_env=KUBE_PATCH_CONVERSION_DETECTOR=true \
                --test_env=KUBE_CACHE_MUTATION_DETECTOR=true \
                -- //... -//vendor/... -//build/...

INFO: Elapsed time: 277.778s, Critical Path: 124.45s
INFO: 2796 processes: 2796 darwin-sandbox.
INFO: Build completed successfully, 2932 total actions
```
After:
```console
bazel test \
                --test_env=KUBE_PATCH_CONVERSION_DETECTOR=true \
                --test_env=KUBE_CACHE_MUTATION_DETECTOR=true \
                -- //... -//vendor/... -//build/...

INFO: Elapsed time: 243.550s, Critical Path: 127.96s
INFO: 2334 processes: 2334 darwin-sandbox.
INFO: Build completed successfully, 2470 total actions
```
This is the same change as https://github.com/atlassian/smith/pull/398.

### DIFF
--- a/cmd/cmd.bzl
+++ b/cmd/cmd.bzl
@@ -6,6 +6,7 @@ def define_command_targets(name, binary_embed):
         name = name,
         embed = binary_embed,
         pure = "on",
+        tags = ["manual"],
         visibility = ["//visibility:public"],
     )
 
@@ -13,7 +14,6 @@ def define_command_targets(name, binary_embed):
         name = name + "_race",
         embed = binary_embed,
         race = "on",
-        tags = ["manual"],
         visibility = ["//visibility:public"],
     )
 
--- a/cmd/crd/BUILD.bazel
+++ b/cmd/crd/BUILD.bazel
@@ -20,6 +20,6 @@ go_library(
 go_binary(
     name = "crd",
     embed = [":go_default_library"],
-    pure = "on",
+    race = "on",
     visibility = ["//visibility:public"],
 )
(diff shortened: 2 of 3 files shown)


ash2k:
Good question!

What gets deployed does not change at all. We deploy race binaries to integration and playground only. Everything else gets pure (statically linked) binaries.

This PR just changes tags on bazel targets. Targets marked with "manual" are not matches by wildcard queries (not sure this is the precise word) - e.g. when you do `bazel test //... -//vendor/...` that means build and test all targets under root except for all targets under `vendor`. This selector will not match `pure="on"` binaries and hence they will not be built. And no libraries that they use will be built. Libraries in pure and race modes must be compiled individually. So we are cutting a lot of work.

What we deploy does not change because we request particular images which request particular binaries. That means CI will still take the same time. This PR significantly reduces dev time though because `make test` is what we use most of the time.

See https://docs.bazel.build/versions/master/guide.html#target-patterns and search for "manual".

> Targets with tags = ["manual"] will not be included in wildcard target patterns ( ..., :*, :all, etc.). You should specify such test targets with explicit target patterns on the command line if you want Bazel to build/test them.

Also see "tags" section here https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes

ash2k:
We have a bunch of integration tests tagged with `manual` and they are not built/ran because of that. To ensure that they can at least be compiled, we build them explicitly. See https://github.com/atlassian/voyager/blob/master/Makefile#L123

PR #37, opened 2019-01-07
Configurable ingress timeouts

brjacobsatlassian:
This adds support for configuring ingress timeouts on Voyager.

It also changes the default from 120s to the micros1 default 60s, and switches over to defaulting via the formation layer (instead of hardcoded into the ingress provider in orchestration).


pkg/orchestration/wiring/internaldns/plugin.go
```diff
@@ -27,15 +27,6 @@ const (
 	kubeIngressRefExample                                   = "ingress-internal-01.ap-southeast-2.paas-dev1.kitt-inf.net"
 )
 
-type Alias struct {
```
brjacobsatlassian: pulled out into their own file because that's the pattern I see with other providers.

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -51,7 +52,7 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
 		return nil, false, errors.Errorf("invalid resource type: %q", resource.Type)
 	}
 
-	deploymentSpec, retriable, err := extractKubeComputeDependency(context.Dependencies)
+	deploymentSpec, retriable, err := extractKubeComputeDependency(&context.Dependencies)
```
brjacobsatlassian: IMO it's a bad idea to pass a variable-sized unbounded object by reference unless the intent is to mutate it. Copying is bad.

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -201,7 +199,42 @@ func buildIngressResource(serviceName smith_v1.ResourceName, resource *orch_v1.S
...
+	if resource.Spec != nil {
+		var rawIngressSpec Spec
+		if err := json.Unmarshal(resource.Spec.Raw, &rawIngressSpec); err != nil {
+			return nil, errors.WithStack(err)
+		}
+
+		if rawIngressSpec.IngressTimeout != nil {
+			timeout = *rawIngressSpec.IngressTimeout
+
+			if timeout < 1 || timeout > 300 {
```
brjacobsatlassian: I don't think this type of validation belongs here but I didn't see where to put it.

pkg/orchestration/wiring/kubeingress/plugin_test.go
```diff
@@ -5,14 +5,94 @@ import (
...
+			},
+		},
+		Exposed: true,
+	}
+
+	emptyStateResource := v1.StateResource{
+		Name: "somename",
+	}
+
+	t.Run("E2E no ingress", func(t *testing.T) {
```
brjacobsatlassian: There was zero unit testing that this thing actually ran, so I wrote one.

pkg/orchestration/wiring/kubeingress/plugin_test.go
```diff
@@ -56,31 +136,39 @@ func TestExtractKubeComputeDependency(t *testing.T) {
 	t.Run("valid single dependency", func(t *testing.T) {
 		deps := []wiringplugin.WiredDependency{computeDep}
 
-		res, _, err := extractKubeComputeDependency(deps)
+		res, _, err := extractKubeComputeDependency(&deps)
 		assert.NoError(t, err)
-		assert.ObjectsAreEqual(deploymentObj, res)
+		assert.Equal(t, deploymentObj, res)
```
brjacobsatlassian: Previous assertion did nothing because ObjectsAreEqual returned a boolean and didn't actually do an assertion

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -232,15 +265,34 @@ func buildIngressHostName(resourceName voyager.ResourceName, sc wiringplugin.Sta
 		clusterHostPath)
 }
 
-func extractKubeComputeDependency(dependencies []wiringplugin.WiredDependency) (*apps_v1.Deployment, bool /* retriable */, error) {
+func extractSingleDependencyOfType(dependencies *[]wiringplugin.WiredDependency, resourceType voyager.ResourceType) (*wiringplugin.WiredDependency, bool /* retriable */, error) {
+	var matchedDependency *wiringplugin.WiredDependency
+	for x := range *dependencies {
```
Burr: you can just do `for _, dependency := range *dependencies` here and then you don't need to reference by index below.

brjacobsatlassian: I copy-pasted this from elsewhere in the codebase (including the index access). Will change, thanks.

brjacobsatlassian: Done.

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -232,15 +265,34 @@ func buildIngressHostName(resourceName voyager.ResourceName, sc wiringplugin.Sta
 		clusterHostPath)
 }
 
-func extractKubeComputeDependency(dependencies []wiringplugin.WiredDependency) (*apps_v1.Deployment, bool /* retriable */, error) {
+func extractSingleDependencyOfType(dependencies *[]wiringplugin.WiredDependency, resourceType voyager.ResourceType) (*wiringplugin.WiredDependency, bool /* retriable */, error) {
```
ash2k: Slice is 2 machine words, don't pass it by pointer

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -232,15 +265,34 @@ func buildIngressHostName(resourceName voyager.ResourceName, sc wiringplugin.Sta
...
+	var matchedDependency *wiringplugin.WiredDependency
+	for _, dependency := range *dependencies {
+		if dependency.Type == resourceType {
+			if matchedDependency != nil {
+				return nil, false, errors.Errorf("must depend on a single %s resource, but multiple were found", resourceType)
+			}
+			matchedDependency = &dependency
```
ash2k: Ha-ha =) This will point to to the loop variable which changes on every iteration.
https://github.com/golang/go/wiki/Range
https://github.com/golang/go/wiki/CommonMistakes related

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -232,15 +265,34 @@ func buildIngressHostName(resourceName voyager.ResourceName, sc wiringplugin.Sta
...
+	}
+
+	return matchedDependency, false, nil
+}
+
+func extractKubeComputeDependency(dependencies *[]wiringplugin.WiredDependency) (*apps_v1.Deployment, bool /* retriable */, error) {
 	// Require exactly one KubeCompute dependency
-	if len(dependencies) != 1 || dependencies[0].Type != apik8scompute.ResourceType {
-		return nil, false, errors.Errorf("must depend on a single %s resource. %d dependencies were given", apik8scompute.ResourceType, len(dependencies))
+	var kubeComputeDependency, retriable, err = extractSingleDependencyOfType(dependencies, apik8scompute.ResourceType)
```
ash2k: Prefer `:=` operator to `var`

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -124,11 +125,9 @@ func buildServiceResource(deploymentName smith_v1.ResourceName, selectorLabels m
 	return serviceResource
 }
 
-// buildIngressResource constructs the Kube / KITT Ingress object
-// with a default rule, plus all aliases rules from dependants internalDNS resources
-func buildIngressResource(serviceName smith_v1.ResourceName, resource *orch_v1.StateResource, context *wiringplugin.WiringContext) (wiringplugin.WiredSmithResource, error) {
+func buildIngressResourceFromSpec(serviceName smith_v1.ResourceName, resourceName voyager.ResourceName, timeout int, context *wiringplugin.WiringContext) (*wiringplugin.WiredSmithResource, error) {
```
ash2k: `serviceName` is not the service name. It is name of the `Service` resource. Just a misleading name. Rename to something like `serviceResourceName` to make it clearer? We use `serviceName voyager.ServiceName` all over the place and this looks like a bug initially.

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -124,11 +125,9 @@ func buildServiceResource(deploymentName smith_v1.ResourceName, selectorLabels m
...
 	var ingressRules []ext_v1b1.IngressRule
-	ingressName := string(resource.Name) + "-" + ingressPostfix
+	ingressName := string(resourceName) + "-" + ingressPostfix
```
ash2k: There is a set of special functions to construct names - `pkg/orchestration/wiring/wiringutil/name.go`

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -124,11 +125,9 @@ func buildServiceResource(deploymentName smith_v1.ResourceName, selectorLabels m
...
 	ingressRuleValue := ext_v1b1.IngressRuleValue{
 		HTTP: &ext_v1b1.HTTPIngressRuleValue{
 			Paths: []ext_v1b1.HTTPIngressPath{
```
ash2k: `ServiceName: string(serviceName),` below is incorrect. Name of the smith resource and name of the kube object, defined by that resource, are not the same thing. Can be, but don't have to be. This is a bug, can be fixed separately.

pkg/orchestration/wiring/kubeingress/plugin.go
ash2k: Validation story is not great ATM, we need to invest more into it. It's on my plate, but too busy atm.

pkg/orchestration/wiring/kubeingress/plugin_test.go
```diff
@@ -5,14 +5,94 @@ import (
...
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+func TestBuildingIngressResource(t *testing.T) {
+	t.Parallel()
+
+	var serviceName smith_v1.ResourceName = "serviceName"
+
+	validOutput := wiringplugin.WiredSmithResource{
```
ash2k: You are using subtest here (which is fine by itself, of course) but I'd rather use just normal tests (lift them up) and a helper function to generate this resource/etc. In general I think it is better to not share state between tests to avoid and potential interference (including caused by bugs in tests) between them.

pkg/orchestration/wiring/kubeingress/plugin_test.go
```diff
@@ -5,14 +5,94 @@ import (
...
+		SmithResource: smith_v1.Resource{
+			Name: "somename-ingress",
```
ash2k: Name of the resource (see below)

pkg/orchestration/wiring/kubeingress/plugin_test.go
```diff
@@ -5,14 +5,94 @@ import (
...
+				},
+			},
+			Spec: smith_v1.ResourceSpec{
+				Object: &ext_v1b1.Ingress{
+					TypeMeta: meta_v1.TypeMeta{
+						Kind:       k8s.IngressKind,
+						APIVersion: ext_v1b1.SchemeGroupVersion.String(),
+					},
+					ObjectMeta: meta_v1.ObjectMeta{
+						Name: "somename-ingress",
```
ash2k: ... and the name of the kube object. They do match here but they don't have to. And things will not work if we change one but not the other. But they should still work.

ash2k (changes requested):

pkg/orchestration/wiring/kubeingress/plugin.go
ash2k: See my comment =)

pkg/orchestration/wiring/kubeingress/plugin.go
brjacobsatlassian: This code has no idea how large "context.Dependencies" is - it's owned by another module. If it contains fifty members, we're going to copy fifty members.

That said, I've changed everything to pass-by-value.

pkg/orchestration/wiring/kubeingress/plugin.go
brjacobsatlassian: Yeah, lovely. Fixed.

pkg/orchestration/wiring/kubeingress/plugin.go
brjacobsatlassian: I didn't change this, and I don't see it as part of this pull request. That said, I have clarified which variable is holding which value.

pkg/orchestration/wiring/kubeingress/plugin_test.go
brjacobsatlassian: I've pulled the fixture setup out into a function.

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -232,19 +266,38 @@ func buildIngressHostName(resourceName voyager.ResourceName, sc wiringplugin.Sta
 		clusterHostPath)
 }
 
+func extractSingleDependencyOfType(dependencies []wiringplugin.WiredDependency, resourceType voyager.ResourceType) (*wiringplugin.WiredDependency, bool /* retriable */, error) {
```
ash2k: boolean can be removed from here

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -232,19 +266,38 @@ func buildIngressHostName(resourceName voyager.ResourceName, sc wiringplugin.Sta
...
+	}
+
+	return matchedDependency, false, nil
+}
+
 func extractKubeComputeDependency(dependencies []wiringplugin.WiredDependency) (*apps_v1.Deployment, bool /* retriable */, error) {
 	// Require exactly one KubeCompute dependency
-	if len(dependencies) != 1 || dependencies[0].Type != apik8scompute.ResourceType {
-		return nil, false, errors.Errorf("must depend on a single %s resource. %d dependencies were given", apik8scompute.ResourceType, len(dependencies))
+	kubeComputeDependency, retriable, err := extractSingleDependencyOfType(dependencies, apik8scompute.ResourceType)
```
ash2k: I didn't realize earlier you are changing the logic here. What is the reason?
We can leave the code as is or we can accept your version - it does not matter because we are refactoring this in autowiring2 (`SmithResources` is going away).

brjacobsatlassian: I didn't see any reason the ingress shouldn't be allowed to depend on any arbitrary resource, so long as there's exactly one kubecompute dependency.

pkg/orchestration/wiring/kubeingress/plugin.go
brjacobsatlassian: Removed.

pkg/orchestration/wiring/kubeingress/plugin.go
ash2k: But what does it mean to `depend on any arbitrary resource`? Why a user might want to do it? The reason to only allow to depend on certain resource types is to prevent user mistakes. They might assume that dependency actually does something when it does not. Or they may have wanted to depend on some other object and used a wrong name by mistake. System will not let them know.

ash2k (approved):
LGTM. Squash merge please.

PR #39, opened 2019-01-08
Print test names in ci mode

mmatusiak-atlassian:
Two things here:
* Print out test names during ci test runs
* Fix the problem with `--config=ci` not being recognized by `bazel query`


.bazelrc
```diff
@@ -10,7 +10,8 @@ build:ci --verbose_failures
 build:ci --curses=no
 build:ci --experimental_sandbox_base=/dev/shm
 
-test:ci --test_output=all
+# Pass -test.v to each test binary to print the test names
+test:ci --test_output=all --test_arg="-test.v"
```
ash2k: Does it work that way? I thought the syntax is 1 flag per line

ash2k: I guess it does https://docs.bazel.build/versions/master/guide.html#bazelrc-syntax-and-semantics

ash2k: However I'd preserve the style and have a flag per line - what we have right now

mmatusiak-atlassian: Yeah that's better, should have done that to begin with.

PR #43, opened 2019-01-08
Shape copy/conversion helper

ash2k:


pkg/orchestration/wiring/ec2compute/common/common.go
```diff
@@ -140,7 +140,12 @@ func WireUp(microServiceNameInSpec, ec2ComputePlanName string, stateResource *or
 	for _, dependency := range dependencies {
 		bound := false
 		if bindableShape, ok := dependency.Contract.FindShape(knownshapes.BindableEnvironmentVariablesShape); ok {
-			resourceReference := bindableShape.(*knownshapes.BindableEnvironmentVariables).Data.ServiceInstanceName
+			typedBindableShape := &knownshapes.BindableEnvironmentVariables{}
```
kopper: This code belongs to knownshapes package, so it is easier for clients to use it. Instead of 

`dependency.Contract.FindShape...` 

I'd prefer

`bindableShapeData := knownshapes.FindBindableEnvironmentVariablesShape(dependency.Contract)`

`FindBindableEnvironmentVariablesShape` should return `BindableEnvironmentVariablesDate`

pkg/orchestration/wiring/wiringplugin/shapes.go
```diff
@@ -39,3 +46,30 @@ type ShapeMeta struct {
...
+		// Same type already, just deep copy
+		sv := reflect.ValueOf(source)
+		tv := reflect.ValueOf(target)
+		method := sv.MethodByName("DeepCopyInto")
+		method.Call([]reflect.Value{tv})
+	default:
+		// Convert by round-tripping to and from JSON
+		data, err := json.Marshal(source)
+		if err != nil {
+			return errors.WithStack(err)
```
kopper: wrap these errors with message to make it more obvious if marshalling or unmarshalling failed

ash2k:
ready for re-review

.bazelrc
```diff
@@ -13,7 +13,6 @@ build:ci --experimental_sandbox_base=/dev/shm
 # Pass -test.v to each test binary to print the test names
 test:ci --test_arg="-test.v"
 test:ci --test_output=all
-test:ci --test_verbose_timeout_warnings
```
brjacobsatlassian: This doesn't look related

pkg/orchestration/wiring/wiringplugin/shapes.go
```diff
@@ -39,3 +46,30 @@ type ShapeMeta struct {
...
+	switch {
+	case st == tt:
+		// Same type already, just deep copy
+		sv := reflect.ValueOf(source)
+		tv := reflect.ValueOf(target)
+		method := sv.MethodByName("DeepCopyInto")
+		method.Call([]reflect.Value{tv})
+	default:
+		// Convert by round-tripping to and from JSON
+		data, err := json.Marshal(source)
```
brjacobsatlassian: Perhaps we should look for a specific converter first?

.bazelrc
ash2k: It's in a separate commit

pkg/orchestration/wiring/wiringplugin/shapes.go
ash2k: What do you mean?

brjacobsatlassian: It'd be nice if there were a way to define a map(from, to) function instead of falling back to serializing and then deserializing to convert any different shapes.

ash2k: Yes, that is how scheme `k8s.io/apimachinery/pkg/runtime.Scheme` works. I thought about it but I think it would be overkill. At least for now we don't need such mechanism, and we can build it later. I've also thought about using `k8s.io/apimachinery/pkg/runtime.DefaultUnstructuredConverter` somehow but there are some potential issues with using it (name of the field in struct not matching the name of the field in JSON) so I stopped looking into it.

PR #45, opened 2019-01-08
ConfigMap with service metadata should be parsed by controller

ash2k:
Entangler does not need to know where the data is coming from. Also support binary data.


pkg/orchestration/controller.go
```diff
@@ -151,6 +157,25 @@ func (c *Controller) process(logger *zap.Logger, state *orch_v1.State) (conflict
 	return conflict, retriable, realBundle, err
 }
 
+func parseConfigMap(configMap *core_v1.ConfigMap) (*orch_meta.ServiceProperties, error) {
+	configMapConfigData, ok := configMap.BinaryData[orch_meta.ConfigMapConfigKey]
+	if !ok {
```
brjacobsatlassian: Kind of interesting to sort it this way - usually you check preconditions, then perform the operation.

This optimizes for the happy path, I guess...

brjacobsatlassian (approved):

PR #46, opened 2019-01-08
Fix ingress resource to respect naming convention

ash2k:
This is not fixing all the identified issues. Will fix in https://github.com/atlassian/voyager/pull/41 (want to avoid conflicts).

### DIFF
--- a/pkg/orchestration/wiring/kubeingress/plugin.go
+++ b/pkg/orchestration/wiring/kubeingress/plugin.go
@@ -40,7 +40,6 @@ const (
 	clusterHostPath = "k8s.atl-paas.net"
 
 	servicePostfix = "service"
-	ingressPostfix = "ingress"
 
 	defaultContourTimeout = 60
 )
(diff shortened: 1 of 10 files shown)


pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -61,30 +60,24 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
 	deploymentName := deploymentSpec.Name
 	deploymentLabels := deploymentSpec.Spec.Template.ObjectMeta.Labels
 
-	var smithResources []wiringplugin.WiredSmithResource
-
 	// Build the Service
-	serviceResource := buildServiceResource(smith_v1.ResourceName(deploymentName), deploymentLabels, resource, context)
-	smithResources = append(smithResources, serviceResource)
+	serviceResource := buildServiceResource(smith_v1.ResourceName(deploymentName), deploymentLabels, resource.Name)
```
ash2k: `smith_v1.ResourceName(deploymentName)` is a bug. Resource name of the resource can be different from the name of the `Deployment` object itself. Not fixing here, as said in the description.

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -61,30 +60,24 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
...
-		Resources: smithResources,
+		Resources: []wiringplugin.WiredSmithResource{serviceResource, ingressResource},
 	}
 
 	return result, false, nil
 }
 
 // buildServiceResource constructs the Kube Service object
-func buildServiceResource(deploymentName smith_v1.ResourceName, selectorLabels map[string]string, resource *orch_v1.StateResource, context *wiringplugin.WiringContext) wiringplugin.WiredSmithResource {
-	serviceName := string(resource.Name) + "-" + servicePostfix
```
ash2k: This is not the right way to construct names.

pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -40,7 +40,6 @@ const (
 	clusterHostPath = "k8s.atl-paas.net"
 
 	servicePostfix = "service"
-	ingressPostfix = "ingress"
```
ash2k: With this PR the ingress resource in Bundle and the Ingress object don't have this postfix.

brjacobsatlassian (approved):

PR #53, opened 2019-01-14
Add anti-affinity and pod disruption budget for kubecompute services

ghost:
This should set sensible defaults to prefer scheduling pods across zones, and then nodes where possible, with a pod disruption budget of 30%.

### DIFF
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -38,6 +38,18 @@ const (
 	WatchVerb            = "watch"
 
 	ServiceDescriptorClaimVerb = "claim"
+
+	// Beta labels, from:
+	// https://github.com/kubernetes/kubernetes/blob/v1.12.2/pkg/kubelet/apis/well_known_labels.go
+	LabelHostname           = "kubernetes.io/hostname"
+	LabelZoneFailureDomain  = "failure-domain.beta.kubernetes.io/zone"
+	LabelMultiZoneDelimiter = "__"
+	LabelZoneRegion         = "failure-domain.beta.kubernetes.io/region"
+
+	LabelInstanceType = "beta.kubernetes.io/instance-type"
+
+	LabelOS   = "beta.kubernetes.io/os"
+	LabelArch = "beta.kubernetes.io/arch"
 )
 
 var (
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -15,6 +15,7 @@ const (
 	DeploymentKind              = "Deployment"
 	NamespaceKind               = "Namespace"
 	PodKind                     = "Pod"
+	PodDisruptionBudgetKind     = "PodDisruptionBudget"
 	ReplicaSetKind              = "ReplicaSet"
 	RoleKind                    = "Role"
 	RoleBindingKind             = "RoleBinding"
--- a/pkg/orchestration/wiring/k8scompute/plugin.go
+++ b/pkg/orchestration/wiring/k8scompute/plugin.go
@@ -3,6 +3,7 @@ package k8scompute
 import (
 	"fmt"
 	"regexp"
+	"sort"
 
 	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
 	"github.com/atlassian/voyager"
@@ -281,9 +282,28 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
 		resourceNameLabel: string(resource.Name),
 	}
 
-	affinity := buildAffinity(resource.Name)
+	affinity := buildAffinity(labelMap)
 
 	podSpec := buildPodSpec(containers, serviceAccountNameRef.Ref(), affinity)
+
+	// Add pod disruption budget
+	pdbSpec := buildPodDisruptionBudgetSpec(labelMap)
+	pdb := smith_v1.Resource{
+		Name: wiringutil.ResourceNameWithPostfix(resource.Name, pdbPostfix),
+		Spec: smith_v1.ResourceSpec{
+			Object: &policy_v1.PodDisruptionBudget{
+				TypeMeta: meta_v1.TypeMeta{
+					Kind:       k8s.PodDisruptionBudgetKind,
+					APIVersion: policy_v1.SchemeGroupVersion.String(),
+				},
+				ObjectMeta: meta_v1.ObjectMeta{
+					Name: wiringutil.MetaNameWithPostfix(resource.Name, pdbPostfix),
+				},
+				Spec: pdbSpec,
+			},
+		},
+	}
+	smithResources = append(smithResources, pdb)
 
 	// The kube deployment object spec
 	deploymentSpec := buildDeploymentSpec(context, spec, podSpec, labelMap, iamRoleRef)
@@ -313,26 +333,6 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
 		Resource: deployment.Name,
 		Path:     metadataNamePath,
 	}
-
-	// Add pod disruption budget
-	pdbSpec := buildPodDisruptionBudgetSpec(resource.Name)
-	pdb := smith_v1.Resource{
-		Name:       wiringutil.ResourceNameWithPostfix(resource.Name, pdbPostfix),
-		References: []smith_v1.Reference{deploymentNameRef},
-		Spec: smith_v1.ResourceSpec{
-			Object: &policy_v1.PodDisruptionBudget{
-				TypeMeta: meta_v1.TypeMeta{
-					Kind:       k8s.PodDisruptionBudgetKind,
-					APIVersion: policy_v1.SchemeGroupVersion.String(),
-				},
-				ObjectMeta: meta_v1.ObjectMeta{
-					Name: wiringutil.MetaNameWithPostfix(resource.Name, pdbPostfix),
-				},
-				Spec: pdbSpec,
-			},
-		},
-	}
-	smithResources = append(smithResources, pdb)
 
 	// 0 value for replicas indicates we don't need an HPA
 	if spec.Scaling.MinReplicas > 0 && spec.Scaling.MaxReplicas > 0 {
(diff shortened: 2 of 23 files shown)


pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -420,6 +445,65 @@ func buildContainers(spec *Spec, envDefault []core_v1.EnvVar, envFrom []core_v1.
...
+				string(resourceName),
+			},
+		},
+	}
+
+	// Create WeightedPodAffinityTerms to configure antiaffinity to distibute
+	// the app to different zones, and then nodes (where possible)
+	podAffinityTerms := []core_v1.WeightedPodAffinityTerm{
+		core_v1.WeightedPodAffinityTerm{
+			Weight: 100,
```
brjacobsatlassian: Using a weight of 100 makes it impossible for us to add another singular rule which trumps this one.

What's the advantage?

pkg/orchestration/wiring/testdata/internaldns.multipleresources.bundle.yaml
```diff
@@ -57,6 +57,27 @@ spec:
                 resourceName: c1
                 stateName: state1
             spec:
+              affinity:
+                podAntiAffinity:
+                  preferredDuringSchedulingIgnoredDuringExecution:
+                  - podAffinityTerm:
+                      labelSelector:
+                        matchExpressions:
+                        - key: app
```
brjacobsatlassian: Discussed offline - fixing to refer to one kubecompute by matching a resource label instead of a service label

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -420,6 +445,65 @@ func buildContainers(spec *Spec, envDefault []core_v1.EnvVar, envFrom []core_v1.
...
+	// Create WeightedPodAffinityTerms to configure antiaffinity to distibute
+	// the app to different zones, and then nodes (where possible)
+	podAffinityTerms := []core_v1.WeightedPodAffinityTerm{
+		core_v1.WeightedPodAffinityTerm{
+			Weight: 75,
+			PodAffinityTerm: core_v1.PodAffinityTerm{
+				LabelSelector: &meta_v1.LabelSelector{
+					MatchExpressions: matchExpressions,
+				},
+				TopologyKey: "failure-domain.beta.kubernetes.io/zone",
```
ash2k: Put this constant into `k8s` package?

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -420,6 +445,65 @@ func buildContainers(spec *Spec, envDefault []core_v1.EnvVar, envFrom []core_v1.
...
+			},
+		},
+		core_v1.WeightedPodAffinityTerm{
+			Weight: 50,
+			PodAffinityTerm: core_v1.PodAffinityTerm{
+				LabelSelector: &meta_v1.LabelSelector{
+					MatchExpressions: matchExpressions,
+				},
+				TopologyKey: "kubernetes.io/hostname",
```
ash2k: Put this constant into `k8s` package?

pkg/orchestration/wiring/testdata/k8scompute.withingressandtimeout.bundle.yaml
```diff
@@ -85,6 +106,22 @@ spec:
...
+      object:
+        apiVersion: policy/v1beta1
+        kind: PodDisruptionBudget
+        metadata:
+          name: kubecompute-simple--pdb
+        spec:
+          minAvailable: 30%
+          selector:
+            matchLabels:
+              resourceName: kubecompute-simple
```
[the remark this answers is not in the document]
ash2k: I think `PodDisruptionBudget` objects are namespaced so we couldn't share a global one anyway (correct me if there is a way, I haven't verified my assumption). And we may want to make them configurable, why not.

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Per discussion, have reduced this to `75`.

pkg/orchestration/wiring/testdata/k8scompute.withingressandtimeout.bundle.yaml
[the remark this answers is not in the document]
halcyonCorsair: Yeah, I checked with KITT, they are namespaced and labelled, _typically_ it would be per compute resource.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -420,6 +445,65 @@ func buildContainers(spec *Spec, envDefault []core_v1.EnvVar, envFrom []core_v1.
...
+func buildAffinity(resourceName voyager.ResourceName) *core_v1.Affinity {
+	return &core_v1.Affinity{
+		PodAntiAffinity: buildAntiAffinity(resourceName),
+	}
+}
+
+func buildAntiAffinity(resourceName voyager.ResourceName) *core_v1.PodAntiAffinity {
+	matchExpressions := []meta_v1.LabelSelectorRequirement{
+		meta_v1.LabelSelectorRequirement{
+			Key:      "app",
```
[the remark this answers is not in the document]
halcyonCorsair: Correct, this should be the `resourceNameLabel`, will push this fix soon.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -47,6 +47,8 @@ const (
 	podSecretEnvVarPluginTypeName = "podsecretenvvar"
 	bindingOutputRoleARNKey       = "IAMRoleARN"
 	envVarIgnoreRegex             = `^IamPolicySnippet$`
+
+	defaultPodDisruptionBudget = "30%"
```
nilebox: We don't seem to use this constant?

pkg/orchestration/wiring/testdata/internaldns.multipleresources.bundle.yaml
halcyonCorsair: Updated to use the resourceName as the key.

pkg/k8s/k8s.go
```diff
@@ -38,6 +38,17 @@ const (
 	WatchVerb            = "watch"
 
 	ServiceDescriptorClaimVerb = "claim"
+
+	// From k8s.io/kubernetes/pkg/kubelet/apis/well_known_labels.go
```
nilebox: Replace it with a GitHub link? https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/well_known_labels.go

Also, there are GA labels in https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/well_known_labels.go#L36-L42
We should probably use them instead of "beta"? We currently run on k8s 1.12, not sure in which version were GA labels released? If they are missing in 1.12, let's create a separate follow-up ticket and leave a TODO here.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -454,7 +456,7 @@ func buildAffinity(resourceName voyager.ResourceName) *core_v1.Affinity {
 func buildAntiAffinity(resourceName voyager.ResourceName) *core_v1.PodAntiAffinity {
 	matchExpressions := []meta_v1.LabelSelectorRequirement{
 		meta_v1.LabelSelectorRequirement{
-			Key:      "app",
+			Key:      resourceNameLabel,
```
nilebox: resource name isn't unique, is it? we should either prefix it with a service name, or use 2 labels (one for service, another for resource)?
/cc @ash2k

nilebox: Just saw the other thread about PodDisruptionBudget being namespaced, then probably it's fine.

ash2k: Resource name is unique within a namespace, yes. We may still want to just use the labels we set on Pods here anyway for consistency. I think that would be good.

pkg/k8s/k8s.go
halcyonCorsair: According to the docs for 1.12 and 1.13, it's still the beta labels: https://v1-12.docs.kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: @ash2k do you mean the resource name label, and the state label from https://github.com/atlassian/voyager/blob/master/pkg/orchestration/wiring/k8scompute/plugin.go#L248-L251 here?

pkg/k8s/k8s.go
halcyonCorsair: And just to be sure, in the source: https://github.com/kubernetes/kubernetes/blob/v1.12.2/pkg/kubelet/apis/well_known_labels.go

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Using it now.

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Done.

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Done.

pkg/k8s/k8s.go
nilebox: @halcyonCorsair update the link with a v1.12.2 one from GitHub?

pkg/orchestration/wiring/k8scompute/plugin.go
ash2k: @halcyonCorsair yes, exactly. Perhaps even pass the map with those labels and transform them into the desired shape here.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -281,6 +287,26 @@ func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext
 		Path:     metadataNamePath,
 	}
 
+	// Add pod disruption budget
+	pdbSpec := buildPodDisruptionBudgetSpec(resource.Name)
+	pdb := smith_v1.Resource{
+		Name:       wiringutil.ResourceNameWithPostfix(resource.Name, pdbPostfix),
+		References: []smith_v1.Reference{deploymentNameRef},
```
ash2k: Why is this reference here? Nothing is needed from Deployment for PDB, right? We don't use a reference for the labels because they are in a different format, but also the order of creation should probably be reversed. PDB should be created first and then the Deployment. Deployment may take time to finish rolling out. But while this is happening we still want to have the PDB constraints be there. Otherwise all pods may end up on a single host =)

halcyonCorsair: The affinity config on the pod handles the spread, the pod disruption budget is simply meant to  prevent us nuking all pods eg. some sort of incident + an upgrade causing too much churn.

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Done.

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Reference has been removed.

ash2k: Sorry, confused two things. So the PDB should either be created before the deployment or independently of it, all good.

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Done.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -447,19 +446,26 @@ func buildContainers(spec *Spec, envDefault []core_v1.EnvVar, envFrom []core_v1.
...
 
-func buildAntiAffinity(resourceName voyager.ResourceName) *core_v1.PodAntiAffinity {
+func buildAntiAffinity(labelMap map[string]string) *core_v1.PodAntiAffinity {
 	matchExpressions := []meta_v1.LabelSelectorRequirement{
 		meta_v1.LabelSelectorRequirement{
 			Key:      resourceNameLabel,
-			Operator: "In",
+			Operator: meta_v1.LabelSelectorOpIn,
+			Values: []string{
+				string(labelMap[resourceNameLabel]),
```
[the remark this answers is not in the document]
halcyonCorsair: Yep, removed.

halcyonCorsair:
Rebased to fix merge conflicts.

PR #54, opened 2019-01-14
MICROS-6602 internaldns webhook

marciogh:
Initial PR for internalDNS webhook.
It should validate existing domain names before Ingress and InternalDNS ServiceInstance creation.
Pending playground tests which I'm running right now, and webhook registration


pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+	return &admissionv1beta1.AdmissionResponse{
+		Allowed: true,
+		Result: &metav1.Status{
+			Message: "allowed domainName",
+		},
+	}, nil
+
+}
+
+func isDomainAllowedToMigrade(ctx context.Context, microsServerClient *microsserver.Client, scClient serviceCentralClient, namespaceService *servicecentral.ServiceData, domainName string) (*admissionv1beta1.AdmissionResponse, error) {
```
mcoot: Typo: `migrade` -> `migrate`

mcoot (dismissed):
LGTM, but get approval from at least one other person before merging, as I haven't written a ValidatingAdmissionWebhook before.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// Validate supported resource type
+	if admissionRequest.Resource != serviceInstanceResource || admissionRequest.Resource != ingressResourceType {
+		return nil, errors.Errorf("unsupported resource, got %v", admissionRequest.Resource)
+	}
+
+	serviceName := getServiceNameFromNamespace(admissionRequest.Namespace)
+	namespaceService, err := getServiceData(ctx, scClient, serviceName)
+	if err != nil {
+		return nil, errors.Errorf("error fetching service central data for namespace %v", admissionRequest.Namespace)
```
Burr: Should use %q for quoted string rather than %v

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+	}, nil
+
+}
+
+func isDomainAllowedToMigrade(ctx context.Context, microsServerClient *microsserver.Client, scClient serviceCentralClient, namespaceService *servicecentral.ServiceData, domainName string) (*admissionv1beta1.AdmissionResponse, error) {
+	aliasInfo, err := microsServerClient.GetAlias(ctx, domainName)
+	if err != nil {
+		return &admissionv1beta1.AdmissionResponse{}, errors.Wrap(err, "error requesting alias info from micros server")
+	}
+	if aliasInfo != (&microsserver.AliasInfo{}) { // response not empty, existing dns alias
```
Burr: this should be a `if aliasInfo != nil` check. The way it is right now (checking against the address of an empty struct)  won't work.

pkg/execution/svccatadmission/util.go
```diff
@@ -14,3 +19,20 @@ func getServiceNameFromNamespace(namespace string) voyager.ServiceName {
 	// DO NOT replicate this approach in other parts of the codebase without discussion.
 	return voyager.ServiceName(strings.Split(namespace, "--")[0])
 }
+
+func getServiceData(ctx context.Context, scClient serviceCentralClient, serviceName voyager.ServiceName) (*servicecentral.ServiceData, error) {
+	search := fmt.Sprintf("service_name='%s'", serviceName)
+	listData, err := scClient.ListServices(ctx, auth.NoUser(), search)
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error looking up service %q", serviceName)
```
Burr: nit: don't capitalize error strings.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+			response, err := isDomainAllowedToMigrade(ctx, microsServerClient, scClient, namespaceService, ingressRule.Host)
+			if err != nil {
+				return nil, err
+			}
+			if response.Allowed == false {
+				return response, nil
+			}
+		}
+	case serviceInstanceResource:
+		serviceInstance := sc_v1b1.ServiceInstance{}
```
Burr: Is there a reason we use a pointer to the `Ingress` above but here we don't? Just a little inconsistent.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/atlassian/voyager/pkg/microsserver"
+	"github.com/atlassian/voyager/pkg/orchestration/wiring/internaldns"
+	"github.com/atlassian/voyager/pkg/servicecentral"
+	sc_v1b1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
+	"github.com/pkg/errors"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	"k8s.io/api/extensions/v1beta1"
```
Burr: might be better to call this `ext_v1beta1` which is what we've tended to use for this package in the past. especially with the other `v1beta1` items already referenced (better to alias them all I think)

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
+package svccatadmission
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/atlassian/voyager/pkg/microsserver"
+	"github.com/atlassian/voyager/pkg/orchestration/wiring/internaldns"
```
Burr: I believe the current practice is to not reference across layers like this. At least, that's not how the other entries in svccatadmission work (e.g. https://github.com/atlassian/voyager/blob/master/pkg/execution/svccatadmission/micros.go#L92). They just validate the plan/service and then have the minimal json to match the expected spec, rather than referencing the whole spec across layers.

If this was discussed in person, please ignore.

ash2k: I think we should change the practice to:
- export the types for the resource type into a subpackage of its autowiring package
- use types from that package.

I have done that change in a very few places already. See `pkg/orchestration/wiring/k8scompute/api` as an example.
We can and should benefit from the fact that we have monorepo but we should not turn our build into "let everything depend on everything and then the linker will prune unused stuff". This would be bad for build speed.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+		for _, alias := range internalDNSSpec.Aliases {
+			response, err := isDomainAllowedToMigrade(ctx, microsServerClient, scClient, namespaceService, alias.Name)
+			if err != nil {
+				return nil, err
+			}
+			if response.Allowed == false {
+				return response, nil
+			}
+		}
+	}
```
ash2k: Should have a `default` case returning a 500 error?

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+		serviceInstance := sc_v1b1.ServiceInstance{}
+		if err := json.Unmarshal(admissionRequest.Object.Raw, &serviceInstance); err != nil {
+			return nil, errors.Wrap(err, "malformed ServiceInstance specification")
+		}
+		var internalDNSSpec internaldns.Spec
+		if err := json.Unmarshal(serviceInstance.Spec.Parameters.Raw, &internalDNSSpec); err != nil {
+			return nil, errors.Wrap(err, "error parsing InternalDNS specs")
+		}
+		for _, alias := range internalDNSSpec.Aliases {
+			response, err := isDomainAllowedToMigrade(ctx, microsServerClient, scClient, namespaceService, alias.Name)
```
ash2k: It would be more efficient to retrieve the whole list in a single call. If this is not possible, we could make multiple concurrent requests. Does not need to be done in this PR or at all, just saying.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+			Message: "allowed domainName",
+		},
+	}, nil
+
+}
+
+func isDomainAllowedToMigrade(ctx context.Context, microsServerClient *microsserver.Client, scClient serviceCentralClient, namespaceService *servicecentral.ServiceData, domainName string) (*admissionv1beta1.AdmissionResponse, error) {
+	aliasInfo, err := microsServerClient.GetAlias(ctx, domainName)
+	if err != nil {
+		return &admissionv1beta1.AdmissionResponse{}, errors.Wrap(err, "error requesting alias info from micros server")
```
ash2k: Should return `nil` as first value.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+						aliasInfo.Service.Name,
+						namespaceService.ServiceName,
+						namespaceService.ServiceOwner.Username,
+					),
+					Code: http.StatusForbidden,
+				},
+			}, nil
+		}
+	}
+	return &admissionv1beta1.AdmissionResponse{}, nil
```
ash2k: It's not great that two functions do this wrapping into `AdmissionResponse`. Make this one return a boolean and construct the `AdmissionResponse` in the caller. That way all `AdmissionResponse` object are constructed in a single function and responsibility for that is centralized vs all over the place.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,103 @@
...
+func InternalDNSAdmitFunc(ctx context.Context, microsServerClient *microsserver.Client, scClient serviceCentralClient, admissionReview admissionv1beta1.AdmissionReview) (*admissionv1beta1.AdmissionResponse, error) {
+
+	admissionRequest := admissionReview.Request
+
+	// Validate supported resource type
+	if admissionRequest.Resource != serviceInstanceResource || admissionRequest.Resource != ingressResourceType {
+		return nil, errors.Errorf("unsupported resource, got %v", admissionRequest.Resource)
+	}
+
+	serviceName := getServiceNameFromNamespace(admissionRequest.Namespace)
```
ash2k: We need to stop using this function and use namespace informers and look at labels instead. But not in this PR of course.

pkg/microsserver/client.go
```diff
@@ -0,0 +1,98 @@
...
+		return httputil.NewUnknown(fmt.Sprintf("%s (%s)", message, http.StatusText(statusCode)))
+	}
+}
+
+func (c *Client) GetAlias(ctx context.Context, domainName string) (*AliasInfo, error) {
+	getAliasEndpoint := "/api/v1/aliases"
+	req, err := c.rm.NewRequest(
+		pkiutil.AuthenticateWithASAP(c.asap, asapAudience, noUser),
+		restclient.Method(http.MethodGet),
+		restclient.JoinPath(fmt.Sprintf(getAliasEndpoint)),
```
ash2k: `fmt.Sprintf`?

ash2k (changes requested):

pkg/execution/svccatadmission/internaldns.go
marciogh: Done in this PR

pkg/execution/svccatadmission/internaldns.go
marciogh: Sorry, this was really bad code. Improved.

marciogh:
Status of this PR: comments addressed, still missing unit tests. Currently testing on playground environment.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,123 @@
...
+					Code: http.StatusForbidden,
+				},
+			}, nil
+		}
+	}
+
+	return &admissionv1beta1.AdmissionResponse{
+		Allowed: true,
+		Result: &metav1.Status{
+			Message: "allowed domainName",
```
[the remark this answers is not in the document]
marciogh: documentation says Result info is never consulted when Allowed:true. Furthermore, that would be a list result (as the request ingress/internaldns can have multiple domains) so I rather keep it simple and rely on code comments for understanding the rule https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/admission/types.go#L90-L93

pkg/execution/svccatadmission/BUILD.bazel
```diff
@@ -27,13 +30,13 @@ go_library(
...
         "//vendor/go.uber.org/zap:go_default_library",
         "//vendor/k8s.io/api/admission/v1beta1:go_default_library",
+        "//vendor/k8s.io/api/extensions/v1beta1:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
     ],
 )
 
 go_test(
     name = "go_default_test",
-    size = "small",
```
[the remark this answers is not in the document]
marciogh: didn't notice that, reverting

pkg/execution/svccatadmission/util.go
```diff
@@ -14,3 +19,20 @@ func getServiceNameFromNamespace(namespace string) voyager.ServiceName {
 	// DO NOT replicate this approach in other parts of the codebase without discussion.
 	return voyager.ServiceName(strings.Split(namespace, "--")[0])
 }
+
+func getServiceData(ctx context.Context, scClient serviceCentralClient, serviceName voyager.ServiceName) (*servicecentral.ServiceData, error) {
+	search := fmt.Sprintf("service_name='%s'", serviceName)
+	listData, err := scClient.ListServices(ctx, auth.NoUser(), search)
```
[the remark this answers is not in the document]
marciogh: I'm not duplicating this code, I'm just moving an existing one from micros.go to util.go. See removed code on pkg/execution/svccatadmission/micros.go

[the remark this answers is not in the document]
nilebox: so then I guess we already have duplication in our code :(

pkg/execution/svccatadmission/internaldns.go
ash2k: Yes, and there is even a ticket to do this.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,132 @@
...
+	// populating a chanel of domains to check owner
+	var domainsToCheck chan string
+	switch admissionRequest.Resource {
+	case ingressResource:
+		ingress := ext_v1b1.Ingress{}
+		if err := json.Unmarshal(admissionRequest.Object.Raw, &ingress); err != nil {
+			return nil, errors.Errorf("error parsing Ingress resource")
+		}
+		if len(ingress.Spec.Rules) == 0 {
+			return nil, errors.Errorf("can not process Ingress with empty rules list")
```
Burr: nit: `cannot` is one word.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,132 @@
...
+				mutex.Unlock()
+			}
+		}()
+	}
+	wg.Wait()
+
+	// checking if domains are allowed to be migrated
+	for _, domainOwnership := range domainsOwnership {
+		if domainOwnership.err != nil {
+			return nil, errors.Wrap(err, fmt.Sprintf("error requesting alias info for %q from micros server", domainOwnership.requestedDomain))
```
Burr: can use `errors.Wrapf`

ash2k (changes requested):

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,132 @@
+package svccatadmission
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/golang/glog"
```
ash2k: Use zap for logging and please don't use the global logger, inject one instead like other components do.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/pkg/errors"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	ext_v1b1 "k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// InternalDNSAdmitFunc checks existing DNS alias ownership via micros server API, for both InternalDNS Services and Ingress Resources
+func InternalDNSAdmitFunc(ctx context.Context, microsServerClient microsServerClient, scClient serviceCentralClient, admissionReview admissionv1beta1.AdmissionReview) (*admissionv1beta1.AdmissionResponse, error) {
+
+	glog.Infof("Received request for %v", admissionReview.Request)
```
ash2k: This information is already available in audit logs.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,132 @@
...
+	type domainOwnership struct {
+		requestedDomain         string
+		existingDomainAliasInfo *microsserver.AliasInfo
+		err                     error
+	}
+	var domainsOwnership []domainOwnership
+
+	// worker groups to parallel fetch requestedDomain owners
+	numWorkers := 5
+	wg := &sync.WaitGroup{}
```
ash2k: `var wg sync.WaitGroup`

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,132 @@
...
+	wg.Add(numWorkers)
+	var mutex = &sync.Mutex{}
```
ash2k: `var mutex sync.Mutex`

pkg/execution/svccatadmission/internaldns_test.go
```diff
@@ -0,0 +1,247 @@
...
+			buildAdmissionResponse(false, http.StatusForbidden, nil, `requested dns alias "elsie.domain" is currently owned by "elsie" via service "elsie-compute-service", and can not be migrated to service "doug-compute-service" owned by different owner "doug"`),
+			false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := InternalDNSAdmitFunc(ctx, microsServerMock, serviceCentralMock, tc.admissionReview)
+			if (err != nil) != tc.wantErr {
+				t.Errorf("InternalDNSAdmitFunc() error = %v, wantErr %v", err, tc.wantErr)
```
ash2k: Replace with `t.Fatalf()` and then you can get rid of `return`.

pkg/execution/svccatadmission/types.go
```diff
@@ -2,7 +2,9 @@ package svccatadmission
 
 import (
 	"context"
+	ext_v1b1 "k8s.io/api/extensions/v1beta1"
```
ash2k: Blank line

pkg/execution/svccatadmission/types.go
```diff
@@ -12,6 +14,7 @@ import (
 const (
 	serviceInstance = "serviceinstances"
 	serviceBinding  = "servicebindings"
+	ingress         = "ingresses"
```
ash2k: These constants can go into `pkg/k8s` package.

pkg/execution/svccatadmission/types.go
```diff
@@ -25,8 +28,17 @@ var (
 		Version:  sc_v1b1.SchemeGroupVersion.Version,
 		Resource: serviceBinding,
 	}
+	ingressResource = metav1.GroupVersionResource{
+		Group:    ext_v1b1.SchemeGroupVersion.Group,
+		Version:  ext_v1b1.SchemeGroupVersion.Version,
+		Resource: ingress,
+	}
```
ash2k: Move these three to `pkg/k8s` package too.

pkg/execution/svccatadmission/util.go
ash2k: Still applies

pkg/microsserver/client.go
```diff
@@ -0,0 +1,98 @@
...
+type Client struct {
+	logger     *zap.Logger
+	httpClient *http.Client
+	asap       pkiutil.ASAP
+	rm         *restclient.RequestMutator
+}
+
+func NewMicrosServerClient(logger *zap.Logger, httpClient *http.Client, asap pkiutil.ASAP, baseURL *url.URL) *Client {
+	rm := restclient.NewRequestMutator(
+		restclient.BaseURL(baseURL.String()),
```
ash2k: Is there a helper that takes a typed URL? Should add one? /cc @scottgreenup

pkg/orchestration/wiring/internaldns/api/types.go
```diff
@@ -0,0 +1,19 @@
+package apiinternaldns
+
+import "github.com/atlassian/voyager"
+
+const (
+	ResourceType                  voyager.ResourceType = "InternalDNS"
+	AliasTypeSimple               string               = "Simple"
+	ClusterServiceClassExternalID                      = "f77e1881-36f3-42ce-9848-7a811b421dd7"
+	ClusterServicePlanExternalID                       = "0a7b1d18-cf8d-461e-ad24-ee16d3da36d3"
```
ash2k: We have types for these ids in `pkg/servicecatalog`

pkg/orchestration/wiring/internaldns/api/types.go
```diff
@@ -0,0 +1,19 @@
+package apiinternaldns
```
ash2k: 👍

pkg/orchestration/wiring/internaldns/api/types.go
```diff
@@ -0,0 +1,19 @@
...
+	ClusterServicePlanExternalID                       = "0a7b1d18-cf8d-461e-ad24-ee16d3da36d3"
+)
+
+type Alias struct {
+	AliasType string `json:"type"`
+	Name      string `json:"name"`
+}
+
+type Spec struct {
+	Aliases []Alias `json:"aliases"`
```
ash2k: If it's ok to have an empty list here then add `omitempty`.

pkg/orchestration/wiring/internaldns/plugin.go
```diff
@@ -17,8 +18,6 @@ import (
 
 const (
 	ResourceType                   voyager.ResourceType = "InternalDNS"
```
ash2k: Remove from here as it is in the `api` subpackage now

ash2k (changes requested):

pkg/execution/svccatadmission/internaldns.go
marciogh: This line was pushed by mistake. Will be fixed on next commits.

pkg/execution/svccatadmission/internaldns.go
marciogh: This line was pushed by mistake. Will be fixed on next commits.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+		var internalDNSSpec apiinternaldns.Spec
+		if err := json.Unmarshal(serviceInstance.Spec.Parameters.Raw, &internalDNSSpec); err != nil {
+			return nil, errors.Wrap(err, "error parsing InternalDNS spec")
+		}
+		if len(internalDNSSpec.Aliases) == 0 {
+			return nil, errors.Errorf("can not process InternalDNS with empty aliases list")
+		}
+		domainsToCheck = make(chan string, len(internalDNSSpec.Aliases))
+		for _, alias := range internalDNSSpec.Aliases {
```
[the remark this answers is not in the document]
marciogh: The location I'm getting the aliases differs between Ingress and ServiceInstance. The first is _ingressRule.Host_ and the second is _alias.Name_. To reuse the code I would have to build a list of domains in a new slice, which seemed to be more clutter than duplicating this little piece of code.

pkg/execution/svccatadmission/internaldns.go
marciogh: Using Sugarred log from context now

pkg/orchestration/wiring/internaldns/api/types.go
marciogh: It's mandatory field

pkg/execution/svccatadmission/util.go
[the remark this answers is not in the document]
marciogh: This (and potentially other duplications) should be solved by merging svccatadmission and computionadmission into a single component.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,129 @@
...
+			defer wg.Done()
+			for domain := range domainsToCheck {
+				aliasInfo, err := microsServerClient.GetAlias(ctx, domain)
+				mutex.Lock()
+				domainsOwnership = append(domainsOwnership, domainOwnership{
+					requestedDomain:         domain,
+					existingDomainAliasInfo: aliasInfo,
+					err:                     err,
+				})
+				mutex.Unlock()
```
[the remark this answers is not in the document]
marciogh: nice, that worked. Will be in the next push

pkg/execution/svccatadmission/util.go
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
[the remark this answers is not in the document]
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
marciogh: fixed

pkg/orchestration/wiring/internaldns/api/types.go
marciogh: fixed

pkg/orchestration/wiring/internaldns/plugin.go
marciogh: fixed

pkg/execution/svccatadmission/types.go
marciogh: fixed

pkg/execution/svccatadmission/types.go
marciogh: fixed. Note that this refactoring added changes to files not related to this PR

pkg/execution/svccatadmission/types.go
marciogh: fixed. Note that this refactoring added changes to files not related to this PR

marciogh:
Status of this PR: all comments so far addressed, manual tests in playground working, still testing edge cases (multiple aliases, multiple domain statuses micros1/2)

pkg/microsserver/client.go
```diff
@@ -0,0 +1,105 @@
...
+		return httputil.NewConflict(message)
+	default:
+		return httputil.NewUnknown(fmt.Sprintf("%s (%s)", message, http.StatusText(statusCode)))
+	}
+}
+
+func (c *Client) GetAlias(ctx context.Context, domainName string) (*AliasInfo, error) {
+
+	logger := logz.RetrieveLoggerFromContext(ctx).Sugar()
+	logger.Infof("fetching alias informtion from micros server for %q", domainName)
```
ash2k: When logging, start with capital letter. When constructing an error message - with normal one.

pkg/microsserver/client.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	defer util.CloseSilently(response.Body)
+	respBody, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read response body")
+	}
+
+	if response.StatusCode == http.StatusNotFound {
+		logger.Infof("alias information for %q not found on micros server", domainName)
```
ash2k: I think this should not be logged here. Same below.

ash2k (changes requested):

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,147 @@
...
+	var mutex = sync.Mutex{}
+	for i := 0; i < numWorkers; i++ {
+		go func() {
+			defer wg.Done()
+			for domain := range domainsToCheck {
+				aliasInfo, err := microsServerClient.GetAlias(ctx, domain)
+				if aliasInfo != nil {
+					logger.Infof("got micros server response %v for domain %v", aliasInfo, domain)
+				}
+				if err != nil {
```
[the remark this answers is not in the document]
marciogh: yeah, this is extra logging I added when debugging an edge  case situation during tests. Removing.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,147 @@
...
+	numWorkers := 5
+	wg := sync.WaitGroup{}
+	wg.Add(numWorkers)
+	var mutex = sync.Mutex{}
+	for i := 0; i < numWorkers; i++ {
+		go func() {
+			defer wg.Done()
+			for domain := range domainsToCheck {
+				aliasInfo, err := microsServerClient.GetAlias(ctx, domain)
+				if aliasInfo != nil {
```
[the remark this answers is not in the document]
marciogh: Here I'm inside a go func, so I can't make the main func return error. What I'm doing instead is stacking both responses (aliasInfo and err) to the domainsOwnership slice, and them checking errors in the main process later, after wg.Wait.

Is there a better way to do that?

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+		}()
+	}
+	wg.Wait()
+
+	// checking if domains are allowed to be migrated
+	for _, domainOwnership := range domainsOwnership {
+		if domainOwnership.err != nil {
+			logger.Errorf("error requesting alias info for %q from micros server", domainOwnership.requestedDomain)
+			return nil, errors.Wrapf(err, "error requesting alias info for %q from micros server", domainOwnership.requestedDomain)
```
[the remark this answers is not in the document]
marciogh: right, extra debugging during tests, removed

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		if domainOwnership.existingDomainAliasInfo != nil {
+			// if we get a domain name with empty Service object from micros-server, it means is a domain provided via OSB for a micros 2 service
+			// in that case micros-server does not have ownership information, so we allow it and let the provider update logic deals with the situation
+			if domainOwnership.existingDomainAliasInfo.Alias.DomainName != "" && domainOwnership.existingDomainAliasInfo.Service.Owner == "" {
+				logger.Infof("domain %q returned no ownership information from micros server, so it is a domain provided via OSB, allowing", domainOwnership.requestedDomain)
```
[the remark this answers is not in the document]
marciogh: As I said before, the documentation says that in case of admission success, the reason is never consulted. So the only way to output messages for success cases is either through logs and comments (I'm doing both)

pkg/microsserver/client.go
marciogh: fixed

pkg/execution/svccatadmission/internaldns.go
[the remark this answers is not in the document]
marciogh: ok, I understood better now. We always have to check error first, before dealing with other returns. Anyway, this whole block is gone, because the checks where here just for debugging logs, and the whole block has been removed now.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,147 @@
...
+	wg := sync.WaitGroup{}
+	wg.Add(numWorkers)
+	var mutex = sync.Mutex{}
+	for i := 0; i < numWorkers; i++ {
+		go func() {
+			defer wg.Done()
+			for domain := range domainsToCheck {
+				aliasInfo, err := microsServerClient.GetAlias(ctx, domain)
+				if aliasInfo != nil {
+					logger.Infof("got micros server response %v for domain %v", aliasInfo, domain)
```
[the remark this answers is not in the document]
marciogh: not needed, removed

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, errors.Errorf("error fetching service central data for serviceName %q", serviceName)
+	}
+
+	// types to store parallel domain owner fetches
+	type domainOwnership struct {
+		requestedDomain         string
+		existingDomainAliasInfo *microsserver.AliasInfo
+		err                     error
+	}
+	var domainsOwnership []domainOwnership
```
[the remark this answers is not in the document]
marciogh: done!

pkg/microsserver/client.go
marciogh: removed unnecessary logging

pkg/execution/svccatadmission/internaldns.go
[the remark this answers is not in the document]
ash2k: It's still here

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, errors.Errorf("error fetching service central data for serviceName %q", serviceName)
+	}
+
+	// types to store parallel domain owner fetches
+	type domainOwnership struct {
+		requestedDomain         string
+		existingDomainAliasInfo *microsserver.AliasInfo
+		err                     error
+	}
+	domainsOwnership := make([]domainOwnership, cap(domainsToCheck))
```
ash2k:
```suggestion
	domainsOwnership := make([]domainOwnership, 0, cap(domainsToCheck))
```
Otherwise you have a non-empty slice.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// worker groups to parallel fetch requestedDomain owners
+	numWorkers := 5
+	wg := sync.WaitGroup{}
```
ash2k:
```suggestion
	var wg sync.WaitGroup
```

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -0,0 +1,141 @@
...
+	wg.Add(numWorkers)
+	var mutex = sync.Mutex{}
```
ash2k:
```suggestion
	var mutex sync.Mutex
```

ash2k (changes requested):
A few minor things

pkg/execution/svccatadmission/internaldns.go
marciogh: done, pushed

pkg/execution/svccatadmission/internaldns.go
marciogh: Done pushed

pkg/execution/svccatadmission/internaldns.go
marciogh: Done pushed

marciogh:
Status update: all cases tested in playground, ready to merge. Related goliath webhook registration https://stash.atlassian.com/projects/KUBE/repos/goliath-ansible/pull-requests/1646/

ash2k (approved):
👍

marciogh:
🏆

PR #55, opened 2019-01-14
VYGR-368: Annotate postgres resource with deletion delay

fcobb:
**Soft Delete Design docs**: https://github.com/atlassian/smith/blob/master/docs/design/soft-deletes.md
**Why:** Prevent unintended deletions of postgres resources

**What:**
- Set postgres deletion delay to one week
- Import time
- Import github.com/atlassian/smith

Signed-off-by: Fraser Cobb <fcobb@atlassian.com>


nilebox:
LGTM, but let's wait for Micros Data team confirmation if they want this change

PR #56, opened 2019-01-16
VYGR-367 - Remove EC2Compute autowiring resource type usage

fcobb:
**Status:** Good to review

This PR removes one usage of ResourceType, more PRs to follow for the remaining usages.


pkg/orchestration/wiring/ec2compute/common/common.go
```diff
@@ -317,3 +308,22 @@ func WireUp(microServiceNameInSpec, ec2ComputePlanName string, stateResource *or
 	return result, false, nil
 
 }
+
+func validateASAPDependencies(context *wiringplugin.WiringContext) error {
+	asapDependencyCount := 0
+	for _, dep := range context.Dependencies {
+		_, found, err := knownshapes.FindASAPKeyShapes(dep.Contract.Shapes)
+		if err != nil {
+			return err
```
scottgreenup: If this becomes the error on the State object in Kubernetes, do we want to wrap this with a useful message for the user?

At the moment it will read 
```
... errors.Wrapf(err, "failed to copy shape %T into %T", shape, copyInto)
```

pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go
```diff
@@ -0,0 +1,43 @@
+package knownshapes
+
+import (
+	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
+	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin"
+)
+
+const (
+	ASAPKeyShapeName wiringplugin.ShapeName = "voyager.atl-paas.net/ASAPKeyShape"
```
scottgreenup: I think we need to drop "Shape" from the end, based on what I see in the other `knownshapes`.
```
ASAPKeyShapeName wiringplugin.ShapeName = "voyager.atl-paas.net/ASAPKey"
```

pkg/orchestration/wiring/wiringutil/knownshapes/util.go
```diff
@@ -1,19 +1,27 @@
...
 		if shape.Name() == name {
+			// Ensure we only have one of the same shape
+			if found {
+				return found, fmt.Errorf("found multiple shapes with name %s", name)
+			}
+			found = true
 			err := wiringplugin.CopyShape(shape, copyInto)
 			if err != nil {
-				return false, errors.Wrapf(err, "failed to copy shape %T into %T", shape, copyInto)
+				return found, errors.Wrapf(err, "failed to copy shape %T into %T", shape, copyInto)
```
scottgreenup: @ychen-atlassian @ash2k @kopper - Just wanted to flag this. We are now returning true instead of false if the shape was found but there was an error copying it.

pkg/orchestration/wiring/wiringutil/knownshapes/zz_generated.deepcopy.go
```diff
@@ -41,6 +66,13 @@ func (in *BindableEnvironmentVariables) DeepCopyShape() wiringplugin.Shape {
 func (in *BindableEnvironmentVariablesData) DeepCopyInto(out *BindableEnvironmentVariablesData) {
 	*out = *in
 	in.BindableShapeStruct.DeepCopyInto(&out.BindableShapeStruct)
+	if in.Vars != nil {
+		in, out := &in.Vars, &out.Vars
+		*out = make(map[string]string, len(*in))
+		for key, val := range *in {
+			(*out)[key] = val
+		}
+	}
```
scottgreenup: @benbarclay This is why we need post commit-hooks :smile:

pkg/orchestration/wiring/ec2compute/common/common.go
```diff
@@ -317,3 +308,22 @@ func WireUp(microServiceNameInSpec, ec2ComputePlanName string, stateResource *or
...
+		}
+
+		if found {
+			// Only allow one asap key dependency per compute
+			// so we can use same micros1 env var names and facilitate migration
+			if asapDependencyCount++; asapDependencyCount > 1 {
+				return errors.Errorf("cannot depend on more than one asap key resource")
```
ash2k: `errors.New()` OR include the number of actual dependencies

pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// +k8s:deepcopy-gen=true
+// +k8s:deepcopy-gen:interfaces=github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin.Shape
+type ASAPKey struct {
+	wiringplugin.ShapeMeta `json:",inline"`
+}
+
+// NewASAPKey creates a new ASAPKey
+func NewASAPKey(resourceName smith_v1.ResourceName) *ASAPKey {
```
ash2k: unused argument

pkg/orchestration/wiring/wiringutil/knownshapes/zz_generated.deepcopy.go
ash2k: We need fuzzing tests for shapes. Like kube has for type structs.

ash2k: Also, `BindableEnvironmentVariables` constructor should have a mandatory parameter for vars to make it a compilation error if that field was not initialized

pkg/orchestration/wiring/asapkey/plugin.go
```diff
@@ -43,11 +45,18 @@ func New() *WiringPlugin {
 			InstanceSpec:                  instanceSpec,
 			ObjectMeta:                    objectMeta,
 			ResourceType:                  ResourceType,
-			OptionalShapes:                svccatentangler.NoOptionalShapes,
+			OptionalShapes:                optionalShapes,
```
nilebox: `Optional` in the field name probably refers to the fact that the implementor doesn't have to set this field if a resource doesn't implement any shapes, i.e. can leave it `nil`? If so, then the function implementing it should just be called `shapes`.

Not sure if my assumption is correct though.

pkg/orchestration/wiring/ec2compute/common/common.go
fcobb: You phrased this as a question, I am not sure if you are just being polite or if you are making a suggestion. Either way I would defer to the teams expertise here.

pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	ASAPKeyShapeName wiringplugin.ShapeName = "voyager.atl-paas.net/ASAPKeyShape"
+)
+
+// +k8s:deepcopy-gen=true
+// +k8s:deepcopy-gen:interfaces=github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin.Shape
+type ASAPKey struct {
+	wiringplugin.ShapeMeta `json:",inline"`
```
nilebox: Shouldn't ASAPKey shape output some data? (ASAP private key etc)? i.e. there should be additional field in the struct.
Shape without data seems useless...

pkg/orchestration/wiring/wiringutil/knownshapes/zz_generated.deepcopy.go
fcobb: IIRC this was generated from `make generate`. Should I remove this section?

pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go
fcobb: @scottgreenup asked @ychen-atlassian annd this is a no for this PR

pkg/orchestration/wiring/asapkey/plugin.go
fcobb: @scottgreenup has a PR open to change this to additional 
https://github.com/atlassian/voyager/pull/58

pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// NewASAPKey creates a new ASAPKey
+func NewASAPKey(_ smith_v1.ResourceName) *ASAPKey {
```
ash2k: Why is this argument still here? =) Why not remove it?

pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go
ash2k: We may need to add data later. e.g. bucket urls.

pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go
fcobb: 👍 thanks, this was an oversight after I copy pasted

PR #57, opened 2019-01-16
Lint DefaultAndValidate in options.go

scottgreenup:
This just seemed strange to me, just quickly reducing code.

### DIFF
--- a/cmd/orchestrationadmission/app/options.go
+++ b/cmd/orchestrationadmission/app/options.go
@@ -14,10 +14,7 @@ type Options struct {
 }
 
 func (o *Options) DefaultAndValidate() []error {
-	var allErrors []error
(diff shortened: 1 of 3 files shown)


cmd/orchestrationadmission/app/options.go
```diff
@@ -14,10 +14,7 @@ type Options struct {
 }
 
 func (o *Options) DefaultAndValidate() []error {
-	var allErrors []error
-	allErrors = append(allErrors, o.ServerConfig.DefaultAndValidate()...)
-
-	return allErrors
+	return o.ServerConfig.DefaultAndValidate()
```
fcobb:
```suggestion
	return "change some code here"
```

PR #58, opened 2019-01-16
Add docs and make optionalShapes consistent in SvcCatEntangler

scottgreenup:
I've added some docs to the SvcCatEntangler. A couple of us took a long time to untangle the entangler and some docs would've helped a lot.

I've also renamed "optional shapes" to "additional shapes". It wasn't clear that it was a function that was adding extra shapes (which in turn implies that there are default shapes). I originally thought it was a way of indicating that these are optional shapes (which is true, but not the entire story).

I've also made `AdditionalShapes` consistent with `References` and `ObjectMeta`, where if any of these are nil they are skipped.

The documentation is also made by looking at the code and asking a few quick questions of @nilebox ; so I'm not sure if each bit of documentation is 100% accurate.

### DIFF
--- a/pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
+++ b/pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
@@ -142,7 +142,7 @@ func (e *SvcCatEntangler) constructResourceContract(resource *orch_v1.StateResou
 	if e.AdditionalShapes != nil {
 		additionalShapes, err := e.AdditionalShapes(resource, smithResource, context)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to compute optional shapes for resource %q of type %q", resource.Name, resource.Type)
+			return nil, errors.Wrapf(err, "failed to compute additional shapes for resource %q of type %q", resource.Name, resource.Type)
 		}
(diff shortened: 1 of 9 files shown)


pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go
```diff
@@ -119,11 +139,13 @@ func (e *SvcCatEntangler) constructResourceContract(resource *orch_v1.StateResou
 	supportedShapes := []wiringplugin.Shape{
 		knownshapes.NewBindableEnvironmentVariables(smithResource.Name, "", nil),
 	}
-	optionalShapes, err := e.OptionalShapes(resource, smithResource, context)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to compute optional shapes for resource %q of type %q", resource.Name, resource.Type)
+	if e.AdditionalShapes != nil {
+		additionalShapes, err := e.AdditionalShapes(resource, smithResource, context)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to compute optional shapes for resource %q of type %q", resource.Name, resource.Type)
```
fcobb:
```suggestion
			return nil, errors.Wrapf(err, "failed to compute additional shapes for resource %q of type %q", resource.Name, resource.Type)
```

PR #62, opened 2019-01-17
MICROS-6895 set alarmEndpoints for OAP and Viceroy

artkoshelev:


pkg/execution/plugins/atlassian/iamrole/iam_role_manager_test.go
```diff
@@ -89,9 +90,21 @@ func TestGenerateRoleInstance(t *testing.T) {
 		ServiceName:     "test-svc-app",
 		OAPResourceName: "app-iamrole",
 		ServiceEnvironment: oap.ServiceEnvironment{
-			NotificationEmail:            "an_owner@example.com",
-			LowPriorityPagerdutyEndpoint: "https://events.pagerduty.com/adapter/cloudwatch_sns/v1/123",
-			PagerdutyEndpoint:            "https://events.pagerduty.com/adapter/cloudwatch_sns/v1/456",
+			NotificationEmail: "an_owner@example.com",
+			AlarmEndpoints: []rds.MicrosAlarmSpec{
```
artkoshelev: we should probably find a better place for `MicrosAlarmSpec` but i failed to find one, so open for suggestions. Maybe `types.go` in a wiring root dir?

pkg/orchestration/wiring/testdata/asapkey.basicdep.bundle.yaml
```diff
@@ -68,9 +68,16 @@ spec:
           computeType: kubeCompute
           oapResourceName: kubecompute-simple-iamrole
           serviceEnvironment:
-            lowPriorityPagerdutyEndpoint: https://events.pagerduty.com/adapter/cloudwatch_sns/v1/12312312312312312312312312312312
+            alarmEndpoints:
```
artkoshelev: actually it's not used in iam role template, so we could just not send them maybe?

pkg/execution/plugins/atlassian/iamrole/iam_role_manager_test.go
marciogh: See related @ash2k suggestion on cross wiring dependencies https://github.com/atlassian/voyager/pull/54#discussion_r247714217  maybe wiring/api ?

ash2k: `wiring` package should not know anything about particular autowiring plugins. I think we can move this type to `pkg/orchestration/wiring/wiringutil/oap/types.go`

artkoshelev:
closing in favor of https://github.com/atlassian/voyager/pull/64

PR #66, opened 2019-01-18
Remove ResourceType usage in internaldns wiring

scottgreenup:
Removed usage of ResourceType from the `internaldns` autowiring function.

FYI: KubeIngress is the only autowiring function that creates an IngressEndpointShape.


pkg/orchestration/wiring/internaldns/plugin.go
```diff
@@ -110,29 +98,32 @@ func mapEnvironmentType(envType voyager.EnvType) string {
 	return string(envType)
 }
 
-// svccatentangler plugin expects reference function to return a slice of reference
-// in this internaldns case, it will always be only one reference, because it's limited by getIngressDependency method
-func getReferences(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) ([]smith_v1.Reference, error) {
+// svccatentangler plugin expects reference function to return a slice of references, in the case of internaldns it will
+// always be a single reference.
+func getReferences(_ *orch_v1.StateResource, context *wiringplugin.WiringContext) ([]smith_v1.Reference, error) {
```
ychen-atlassian: Why not just remove the resource argument given this is a private function?

PR #67, opened 2019-01-20
Scaffolding for status autowiring

ash2k:
This refactoring is needed to allow autowiring functions define status for their resources. In this PR we introduce the possibility of doing it, without actually touching the functions yet. An adapter is used to augment existing functions with legacy behavior to preserve current output for now.
Next step is to modify each function to implement the new `Status()` method natively and eventually get rid of the wrapper.
Most of the functions use `SvcCatEntangler` which can be changed to implement this new function in a specific way.


ash2k:
Ready for review

PR #68, opened 2019-01-21
NONE: Run make generate to update deployinator

fcobb:
**Status:** Ready
**What/Why:**
- I ran 'make generate' on master and these were the results.
- Should we commit these changes?

PR #69, opened 2019-01-21
Fix generate target to run gazelle and goimports

ash2k:
Fix cause of issues discovered in https://github.com/atlassian/voyager/pull/68.


ash2k:
It also generates all these sets again...

PR #70, opened 2019-01-21
VYGR-367: Remove resource type usage in Postgres wiring plugin

fcobb:
**What:**

- Add RDS shape with name only (no data)
- Output RDS shape in RDS plugin
- Use RDS shape in Postgres wiring


pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -72,22 +72,32 @@ func New() *WiringPlugin {
...
+	if len(rdsDependency) == 0 {
+		return wiringplugin.WiredDependency{}, nil
+	}
+
+	// Ensure postgres does not depend on more than one RDS resource
+	if len(rdsDependency) > 1 {
+		return wiringplugin.WiredDependency{}, errors.Errorf("postgres resources can only depend on one RDS resource type")
+	}
+
+	return rdsDependency[0], nil
```
fcobb: I reworked this function a little bit as I find it simple to have the happy case as the "first tab" and error/unexpected cases handled in the "second tab" aka indent path.

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -72,22 +72,32 @@ func New() *WiringPlugin {
...
-	return wiringplugin.WiredDependency{}, nil
+
+	// Did not find a RDS dependency
+	if len(rdsDependency) == 0 {
+		return wiringplugin.WiredDependency{}, nil
+	}
+
+	// Ensure postgres does not depend on more than one RDS resource
+	if len(rdsDependency) > 1 {
+		return wiringplugin.WiredDependency{}, errors.Errorf("postgres resources can only depend on one RDS resource type")
```
fcobb: Removed the capitalized `Postgres`

pkg/orchestration/wiring/wiringutil/knownshapes/rds.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+// Name returns the RDS ShapeName
+func (a *RDS) Name() wiringplugin.ShapeName {
+	return a.ShapeName
+}
+
+// FindRDSShapes returns a single instance of the RDS shape if found and will error if there are multiples
+func FindRDSShapes(shapes []wiringplugin.Shape) (*RDS, bool /*found*/, error) {
```
fcobb: This is a copy-pasta from the ASAP shape: https://github.com/atlassian/voyager/blob/master/pkg/orchestration/wiring/wiringutil/knownshapes/asap_key.go#L32-L42

pkg/orchestration/wiring/wiringutil/knownshapes/standard_test.go
```diff
@@ -31,6 +31,7 @@ func TestAllKnownShapes(t *testing.T) {
 		NewSetOfPodsSelectableByLabels(resourceName, map[string]string{"a": "b"}),
 		NewSnsSubscribable(resourceName),
 		NewASAPKey(),
+		NewRDS(),
```
fcobb: Do we want these alphabetized?

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -72,22 +72,32 @@ func New() *WiringPlugin {
 	}
 }
 
+// getRDSDependency returns a single RDS dependency if found in a given list, returning an error if multiple are found
 func getRDSDependency(dependencies []wiringplugin.WiredDependency) (wiringplugin.WiredDependency, error) {
```
ash2k: Use https://github.com/atlassian/voyager/blob/13530e71e9fb03823878012ae88316d52c52c9d4/pkg/orchestration/wiring/wiringplugin/types.go#L33 to get the only dependency and then look for the needed shape in it.

pkg/orchestration/wiring/wiringutil/knownshapes/standard_test.go
ash2k: Doesn't matter, but wouldn't hurt.

ychen-atlassian:
FYI in my PR I've added the shape as well so we'll run into conflicts. I've named the shape "SharedDbShape". In particular it returns some extra data ("ReadReplica") which is necessary for the Postgres autowiring. Feel free to rename.

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -72,38 +72,33 @@ func New() *WiringPlugin {
...
-	}
-	return wiringplugin.WiredDependency{}, nil
-}
-
 func references(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) ([]smith_v1.Reference, error) {
 	references := []smith_v1.Reference{}
-	rdsDependency, err := getRDSDependency(context.Dependencies)
+	dep, found, err := context.FindTheOnlyDependency()
+	if err != nil {
+		return references, err
```
ash2k: Return `nil` instead of `references` here and below

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -72,38 +72,33 @@ func New() *WiringPlugin {
...
+
+	// Found dependency but it was not a RDS resource
+	if !found {
 		return references, nil
 	}
-	instanceName := wiringutil.ServiceInstanceResourceName(rdsDependency.Name)
+
+	instanceName := wiringutil.ServiceInstanceResourceName(dep.Name)
 	referenceName := wiringutil.ReferenceName(instanceName, "metadata-name")
 	references = append(references, smith_v1.Reference{
```
ash2k: Just construct the slice using a literal right here, no need for a variable

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -164,13 +159,26 @@ func instanceSpec(resource *orch_v1.StateResource, context *wiringplugin.WiringC
...
+	// Did not find any dependencies
+	if !found {
+		return json.Marshal(finalSpec)
+	}
+
+	// Check if dependency has a RDS shape
+	_, found, err = knownshapes.FindRDSShapes(dep.Contract.Shapes)
 	if err != nil {
-		return nil, err
+		return json.Marshal(finalSpec)
```
ash2k: This should return an error if there IS a dependency but something is wrong with its shape

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -72,38 +72,33 @@ func New() *WiringPlugin {
...
-			return wiringplugin.WiredDependency{}, errors.Errorf("Postgres resources can only depend on one RDS resource type")
-		}
-		if len(rdsDependency) == 1 {
-			return rdsDependency[0], nil
-		}
-	}
-	return wiringplugin.WiredDependency{}, nil
-}
-
 func references(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) ([]smith_v1.Reference, error) {
```
ash2k: This entangler needs some refactoring. It is silly that we need to do this dependency extracting twice. I'll do it later.

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -164,13 +159,26 @@ func instanceSpec(resource *orch_v1.StateResource, context *wiringplugin.WiringC
...
+	// Check if dependency has a RDS shape
+	_, found, err = knownshapes.FindRDSShapes(dep.Contract.Shapes)
 	if err != nil {
-		return nil, err
+		return json.Marshal(finalSpec)
 	}
-	if rdsDependency.Name != "" {
+
+	// Found RDS dependency
+	if found {
```
ash2k: Shouldn't there be an `else` here that returns an error?

pkg/orchestration/wiring/wiringplugin/types.go
```diff
@@ -30,7 +30,8 @@ type WiringContext struct {
 	Dependants   []DependantResource
 }
 
-func (c *WiringContext) TheOnlyDependency() (*WiredDependency, error) {
+// GetTheOnlyDependency will return a single dependency, returning an error if there is more or less than one
+func (c *WiringContext) GetTheOnlyDependency() (*WiredDependency, error) {
```
ash2k: `Get...` is a Java convention, not a thing in Go. Please rename back.

ash2k:
There is a merge conflict...

pkg/orchestration/wiring/postgres/plugin.go
fcobb:
```suggestion
		return nil, errors.Wrapf(err, "unable to determine if shape %s db was a dependency", knownshapes.SharedDbShape)
```

fcobb: Does the suggestion above meet your expectations?

pkg/orchestration/wiring/testdata/postgres.dependsons3.bundle.yaml
```diff
@@ -1,50 +0,0 @@
-apiVersion: smith.atlassian.com/v1
```
fcobb: After talking to @ash2k @nilebox and @marciogh we decided that the state should output and error, therefore I replaced this bundle with an error

PR #71, opened 2019-01-21
Kubernetes 1.13 libraries

ash2k:
See individual commits.


brjacobsatlassian:
Conflicted

vendor/k8s.io/cluster-registry/pkg/client/clientset/versioned/typed/clusterregistry/v1alpha1/fake/fake_cluster.go
```diff
@@ -130,8 +130,9 @@ func (c *FakeClusters) DeleteCollection(options *v1.DeleteOptions, listOptions v
 
 // Patch applies the patch and returns the patched cluster.
 func (c *FakeClusters) Patch(name string, pt types.PatchType, data []byte, subresources ...string) (result *v1alpha1.Cluster, err error) {
+	// EDITED MANUALLY TO MAKE IT COMPILE https://github.com/kubernetes/cluster-registry/issues/271
```
brjacobsatlassian: Do we really want to do this?

Makefile
```diff
@@ -424,7 +424,6 @@ generate-deepcopy:
 	find pkg -name zz_generated.deepcopy.go -delete
 	bazel build $(BAZEL_OPTIONS) //vendor/k8s.io/code-generator/cmd/deepcopy-gen
 	./bazel-bin/vendor/k8s.io/code-generator/cmd/deepcopy-gen/$(BINARY_PREFIX_DIRECTORY)/deepcopy-gen $(VERIFY_CODE) \
-	--v 1 --logtostderr \
```
brjacobsatlassian: Not related to this change?

ash2k:
I have switched to a fork instead of modifying vendor directly.

brjacobsatlassian:
I think we should maintain this as an explicit set of patches, rather than a forked repository. That improves the impetus to "go back to stock" and makes our changes more explicit, IMO.

ash2k:
> I think we should maintain this as an explicit set of patches, rather than a forked repository.

I think this would be more work and is not worth it for a 1 line change.

ash2k:
Ready for review

cmd/composition/app/main.go
```diff
@@ -12,10 +12,12 @@ import (
 	"github.com/atlassian/voyager/cmd"
 	"github.com/atlassian/voyager/pkg/composition"
 	"github.com/atlassian/voyager/pkg/util/crash"
+	"k8s.io/klog"
 )
 
 func Main() {
 	rand.Seed(time.Now().UnixNano())
+	klog.InitFlags(nil)
```
nilebox: So we do need to explicitly call `klog.InitFlags` but shouldn't for all API servers (to avoid conflict with API server code), correct?

nilebox: would be nice to raise an issue about this in k/k repo

ash2k: yes to both

PR #72, opened 2019-01-22
MICROS-6895 set alarmEndpoints for EC2 resource

artkoshelev:


pkg/orchestration/wiring/testdata/ec2compute.overwrite.state.yaml
```diff
@@ -20,9 +20,6 @@ spec:
         maxSize: 2
       notifications:
         email: abc@email.com
-        pagerduty:
```
artkoshelev: we are not going to support env overrides for alarms since it would be solved in OpsGenie alarm routing rules

ash2k (approved):

PR #73, opened 2019-01-22
Make OSB wiring easier to read

nilebox:
Implemented a set of lighter OSB helper functions as an alternative to `SvcCatEntangler`, which IMO makes reading the code confusing due to 4 callback functions (it's hard to understand what each of these functions do unless you read the code in `SvcCatEntangler`).

Migrated `Postgres` resource on the new OSB helper code to demonstrate the difference.

### DIFF
--- a/pkg/orchestration/wiring/wiringutil/osb/service_instance_wiring.go
+++ b/pkg/orchestration/wiring/wiringutil/osb/service_instance_wiring.go
@@ -1,0 +1,57 @@
+package osb
+
+import (
+	"encoding/json"
+
+	orch_v1 "github.com/atlassian/voyager/pkg/apis/orchestration/v1"
+	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringutil"
+	"github.com/atlassian/voyager/pkg/servicecatalog"
+	sc_v1b1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
+	"github.com/pkg/errors"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type partialSpec struct {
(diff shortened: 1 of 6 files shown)


pkg/orchestration/wiring/wiringutil/osb/service_instance_wiring.go
```diff
@@ -0,0 +1,72 @@
+package osb
```
nilebox: This code is based on https://github.com/atlassian/voyager/blob/master/pkg/orchestration/wiring/wiringutil/svccatentangler/plugin.go, but re-writen using a set of helper functions rather than a whole wiring plugin implementation with callbacks

pkg/orchestration/wiring/wiringutil/osb/service_instance_wiring.go
```diff
@@ -0,0 +1,57 @@
...
+// If the spec is empty or does not contain instance ID, then this returns the empty string.
+func instanceID(resourceSpec *runtime.RawExtension) (string, error) {
+	if resourceSpec == nil {
+		return "", nil
+	}
+
+	spec := partialSpec{}
+	err := json.Unmarshal(resourceSpec.Raw, &spec)
+	if err != nil {
+		return "", errors.Wrap(err, "unmarshalling StateResource failed")
```
ash2k: invalid comment?

pkg/orchestration/wiring/wiringutil/osb/service_instance_wiring.go
brjacobsatlassian: One thing to watch out for is that helper libraries can grow out of control pretty quickly :-/

nilebox: @brjacobsatlassian at the moment it's opposite (see original code), but we should keep an eye on this code, of course.

PR #74, opened 2019-01-22
EDGE-891: global edge provider integration

cetanu:


roaanv:
I think these might help:
https://hello.atlassian.net/wiki/spaces/MDEV/pages/384865916/Global+Edge+integration+for+KubeCompute

To get the FQDN that was set in the ServiceDescriptor: (although a, as yet not yet built, ExternalDNSResources needs to be used):
https://github.com/atlassian/voyager/blob/dec177b6d3fa972407dcb97542b920db89eb76b0/pkg/orchestration/wiring/kubeingress/plugin.go#L144

To get the address of the KITT edge:
https://github.com/atlassian/voyager/blob/dec177b6d3fa972407dcb97542b920db89eb76b0/pkg/orchestration/wiring/internaldns/plugin.go#L111

cetanu:
The diff should be looking a lot better now

pkg/orchestration/wiring/edge/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+type WiringPlugin struct {
+}
+
+func New() *WiringPlugin {
+	return &WiringPlugin{}
+}
+
+func (p *WiringPlugin) WireUp(resource *orchestration.StateResource, context *wiring.WiringContext) (result *wiring.WiringResult, retriable bool, err error) {
```
ash2k: Please don't use named return values and naked returns. They are a source of bugs. Also for codebase style consistency.

pkg/orchestration/wiring/edge/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	// Pass-through SD parameters -> OSB attributes
+	var attributes Attributes
+	if err := json.Unmarshal(resource.Spec.Raw, &attributes); err != nil {
+		return nil, errors.WithStack(err)
+	}
+	parameters := InstanceParameters{
+		ServiceName: string(context.StateContext.ServiceName),
+		Resource: ResourceParameters{
```
ash2k: Is this nesting necessary? i.e. it'd look better with less nesting.

pkg/orchestration/wiring/edge/types.go
```diff
@@ -2,15 +2,24 @@ package edge
...
+	UpstreamPort    int              `json:"upstream_port,omitempty"`
+	UpstreamSuffix  string           `json:"upstream_suffix,omitempty"`
+	UpstreamOnly    string           `json:"upstream_only,omitempty"`
+	Domain          []string         `json:"domain,omitempty"`
+	Healthcheck     string           `json:"healthcheck,omitempty"`
+	Rewrite         string           `json:"rewrite,omitempty"`
+	Routes          Routes           `json:"routes,omitempty"`
 }
 
 type UpstreamAddress []struct {
```
roaanv: OOC: What does this mean? I.e. how does this type work.
```
x := UpstreamAddress{}
x[0] = ???
````

File: pkg/orchestration/wiring/edge/plugin.go [hunk 4]
```diff
@@ -2,52 +2,91 @@ package edge
...
+			Attributes: attributes,
+		},
+
+	}
+
+	if parameters.Resource.Attributes.UpstreamAddress == nil {
+		return nil, errors.New("UpstreamAddress is required")
+
+		// TODO: Make upstream address optional and produce it from InternalDNS / KubeIngress output
```
roaanv: This will have to be from EC2Compute and KubeIngress.
NOT InternalDNS

pkg/orchestration/wiring/edge/types.go
nilebox: Interesting, never saw this pattern before :)

@roaanv type is just an alias for a particular structure, e.g. you can define `type Bla string` which will operate the same way as a normal string. Here a type is defined as "slice of anonymous structs".
> x := UpstreamAddress{}

it's the same as
> x := []struct{}

i.e. it's an empty slice and `x[0]` will panic, see https://play.golang.org/p/mkYe8VkxR5i

nilebox:
```suggestion
 type UpstreamAddress struct {
```

pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+	ServiceName string             `json:"serviceName"`
+	Resource    ResourceParameters `json:"resource"`
+}
+
+type ResourceParameters struct {
+	Attributes Attributes `json:"attributes"`
+}
+
+type Attributes struct {
+	UpstreamAddress *UpstreamAddress `json:"upstream_address"`
```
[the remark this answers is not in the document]
nilebox:
```suggestion
	UpstreamAddress []UpstreamAddress `json:"upstream_address"`
```

pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+	UpstreamPort    int              `json:"upstream_port,omitempty"`
```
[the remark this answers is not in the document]
nilebox:
```suggestion
	UpstreamPort    int32            `json:"upstream_port,omitempty"`
```

pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+	Route    RouteAction `json:"route,omitempty"`
+	Redirect string      `json:"redirect,omitempty"`
+}
+
+type RouteMatch struct {
+	Prefix string `json:"prefix,omitempty"`
+	Regex  string `json:"regex,omitempty"`
+	Path   string `json:"path,omitempty"`
+	Host   string `json:"host,omitempty"`
+}
```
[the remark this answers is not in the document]
nilebox: to clarify: `omitempty` leads to not marshalling primitive fields with default values. e.g. in case of strings if `Host == ""`, it won't be present in JSON after marshalling.

pkg/orchestration/wiring/testdata/edge.provided_upstream.state.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: orchestration.voyager.atl-paas.net/v1
```
nilebox: Looks good, just add some more examples of state.yaml with validation errors (e.g. empty resource spec or empty `upstream_address: []`.

pkg/orchestration/wiring/edge/plugin.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	parameters := InstanceParameters{
+		ServiceName: string(context.StateContext.ServiceName),
+		Resource: ResourceParameters{
+			Attributes: attributes,
+		},
+
+	}
+
+	if parameters.Resource.Attributes.UpstreamAddress == nil {
```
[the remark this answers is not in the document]
nilebox:
```suggestion
	if len(parameters.Resource.Attributes.UpstreamAddress) == 0 {
```

pkg/orchestration/wiring/edge/types.go
nilebox: Givent that the struct is anonymous, appending items to this slice is very ugly (but works!) https://play.golang.org/p/DDLXHxTKqT8

pkg/orchestration/wiring/edge/types.go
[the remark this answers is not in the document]
cetanu: Yep this is intentional, I think the desired experience was for the user to not have to input anything, and to have most benefits by default

pkg/orchestration/wiring/edge/plugin.go
cetanu: Removed naked returns from this

pkg/orchestration/wiring/edge/plugin.go
cetanu: Updated

pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Resource    ResourceParameters  `json:"resource"`
+}
+
+type ResourceParameters struct {
+	Attributes Attributes `json:"attributes"`
+}
+
+type Attributes struct {
+	UpstreamAddress []UpstreamAddress `json:"upstream_address"`
```
nilebox: formatting is off there (Go has strict formatting rules), just run `make fmt` and commit the change

nilebox:
The code looks good to me, just add some unhappy test scenarios

cetanu:
Added some test scenarios, have to fix one of them... It seems that it's adding an empty struct to the bundle, not sure why. If you run the tests you'll see what I mean

nilebox:
The builds don't run because it's a pull request from a fork, not within the same repository :(
@ychen-atlassian would it be hard to support forked repos as well?

ash2k:
We shouldn't run builds with code from arbitrary PRs from arbitrary forks. Instead we should give people proper permissions for this repo. It's easier to work that way anyways.

nilebox:
> We shouldn't run builds with code from arbitrary PRs from arbitrary forks. Instead we should give people proper permissions for this repo.

We won't be giving these permissions to people outside Atlassian, so we should then lock down PRs to the members of the same Atlassian group as we configured for pushing branches to this repo. Also, "Pull requests, issues and comments welcome" in https://github.com/atlassian/voyager/blob/master/README.md is misleading, it's better to call out that pull requests are not "welcome" from people outside Atlassian.

nilebox:
Pushed the branch to the Voyager repo and raised a new pull request #162
@cetanu please continue work there.

PR #77, opened 2019-01-23
VYGR-373 iampolicysnippet plugin refactor to not read dependencies

ychen-atlassian:
This change aims to reduce the need to read the plugin dependencies by having smith perform the injection of secrets into the spec (actually this raises of question of if plugins should even need the dependencies passed into it?)

As usual, the commits are split into:

1. Tests
2. Modification of the iamrole smith plugin
3. Modification of resources to return the correct IamPolicy shape

I tested this on the E2E Service Descriptor and the IAM policy has not changed since I provisioned the service last week. (produces exactly the same policy)

image

I've also confirmed that the updated ServiceInstance contains the snippets inside the spec. I don't believe there should be any confident information in an IamPolicySnippet? Theoretically this is also "safer" because the user can't just mess with the secrets containing the snippets and grant themselves access to everything (because we block ServiceInstance mutations, for now)


pkg/orchestration/wiring/wiringutil/iam/iam.go
```diff
@@ -28,31 +29,53 @@ const (
...
+
+	// This is just the local reference name
+	dependencyNamePostfix = "iamrole"
 )
 
-func PluginServiceInstance(computeType iam_plugin.ComputeType, resourceName voyager.ResourceName, serviceName voyager.ServiceName, createInstanceProfile bool, dependencyReferences []smith_v1.Reference,
+type ResourceWithIamAccessibleBinding struct {
+	ResourceName               voyager.ResourceName
+	BindableIamAccessibleShape knownshapes.BindableIamAccessible
+	CreatedBindingFromShape    smith_v1.Resource
```
ash2k: If only the name is needed it would be much more efficient to only have the name here. Same with the other similar struct for env vars.

pkg/orchestration/wiring/wiringutil/iam/iam.go
```diff
@@ -28,31 +29,53 @@ const (
...
+}
+
+func PluginServiceInstance(computeType iam_plugin.ComputeType, stateResourceName voyager.ResourceName,
+	serviceName voyager.ServiceName, createInstanceProfile bool, iamShapedResources []ResourceWithIamAccessibleBinding,
 	context *wiringplugin.WiringContext, managedPolicies, assumeRoles []string) (smith_v1.Resource, error) {
 
+	dependencyReferences := []smith_v1.Reference{}
```
ash2k: Can pre-allocate, and below

ash2k:
I've commented only on one of the compute resoruce, but same comments apply to both.

pkg/orchestration/wiring/ec2compute/common/common.go
```diff
@@ -115,46 +115,60 @@ func WireUp(microServiceNameInSpec, ec2ComputePlanName string, stateResource *or
...
-	for _, res := range bindingResult {
-		dependencyReferences = append(dependencyReferences, smith_v1.Reference{
-			Resource: res.CreatedBindingFromShape.Name,
-		})
+		// We also depend on BindableIamAccessible shape
+		bindableIamAccessibleShape, iamFound, err := knownshapes.FindBindableIamAccessibleShape(dependency.Contract.Shapes)
+		if err != nil {
+			return nil, false, err
+		}
+		if iamFound {
```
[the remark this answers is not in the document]
ychen-atlassian: I left this as `if iamFound` such that later if we have more shapes it's a lot easier to understand what is happening.

```
_, found, _ := find()
if found {
   xxx
}

_, found, _ := findAnotherShape()
if !found {
...
}
```

E.g. something like this might happen in the future and it seems more consistent if everything checks a particular branch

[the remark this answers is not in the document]
ychen-atlassian: Then again I don't really care too much so I'll just change it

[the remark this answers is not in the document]
ychen-atlassian: I changed it just to test and tbh it looks a lot messier outside of the if-statement

While there is indenting I think the fact that the code is surrounded by braces makes it easier to see what code belongs to which section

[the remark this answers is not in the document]
ash2k: Yeah, I see your point. Keep it as is, I agree.

pkg/execution/plugins/atlassian/iamrole/iam_role_manager.go
```diff
@@ -169,18 +169,18 @@ func defaultJSON(serviceName voyager.ServiceName) *IamPolicy {
 }
 
 func generateRoleInstance(spec *Spec) (*sc_v1b1.ServiceInstance, error) {
-	policyDocuments := []*IamPolicyDocument{}
+	policyDocuments := make([]*IamPolicyDocument, 0, len(spec.PolicySnippets))
```
ash2k: This shouldn't be a pointer - you need to allocate all this memory anyway. Allocating in a single block is more efficient.

ash2k:
I have pushed a commit to remove some pointers, please have a look.

pkg/execution/plugins/atlassian/iamrole/iam_role_manager.go
```diff
@@ -171,10 +168,19 @@ func defaultJSON(serviceName voyager.ServiceName) *IamPolicy {
...
 
-func generateRoleInstance(spec *Spec, dependencies map[smith_v1.ResourceName]smith_plugin.Dependency) (*sc_v1b1.ServiceInstance, error) {
-	policyDocuments, err := extractIamPolicies(dependencies)
-	if err != nil {
-		return nil, err
+func generateRoleInstance(spec *Spec) (*sc_v1b1.ServiceInstance, error) {
+	policyDocuments := make([]IamPolicyDocument, 0, len(spec.PolicySnippets))
+	for resourceName, policySnippet := range spec.PolicySnippets {
+		policy := IamPolicyDocument{}
+		if err := json.Unmarshal([]byte(policySnippet), &policy); err != nil {
```
brjacobsatlassian: Was policySnippet a string? Why don't we have a string unmarshaller?

pkg/execution/plugins/atlassian/iamrole/iam_role_manager.go
```diff
@@ -328,69 +306,49 @@ func makeStatementKey(statement *IamPolicyStatement) (string, error) {
...
+				if !present {
+					statementByKey = IamPolicyStatement{
 						Effect:    "Allow",
 						Action:    statement.Action,
 						NotAction: statement.NotAction,
 						Condition: statement.Condition,
 					}
 				}
-				statementsByKey[statementKey].Resource = append(statementsByKey[statementKey].Resource, resource)
+				statementByKey.Resource = append(statementByKey.Resource, resource)
```
brjacobsatlassian: This is appending a resource to a thing called "Resource". That makes no sense - should be "Resources" if it's a slice/array.

pkg/execution/plugins/atlassian/iamrole/testdata/badpolicy.error
```diff
@@ -1 +1 @@
-failed to extract IAM policy from the resource "dynamo": failed to unmarshal IAM policy from JSON object: invalid character 'm' looking for beginning of value
+failed to extract IAM policy from the resource "dynamo": invalid character 'm' looking for beginning of value
```
brjacobsatlassian: I hope we never expose this no-position-information parse error to users >_<

pkg/orchestration/wiring/ec2compute/common/common.go
```diff
@@ -115,46 +115,60 @@ func WireUp(microServiceNameInSpec, ec2ComputePlanName string, stateResource *or
...
-	}
 
-	// filter out the list of dependencyReferences - this will be used
-	dependencyReferences := make([]smith_v1.Reference, 0, len(bindingResult))
-	for _, res := range bindingResult {
-		dependencyReferences = append(dependencyReferences, smith_v1.Reference{
-			Resource: res.CreatedBindingFromShape.Name,
-		})
+		// We also depend on BindableIamAccessible shape
+		bindableIamAccessibleShape, iamFound, err := knownshapes.FindBindableIamAccessibleShape(dependency.Contract.Shapes)
```
brjacobsatlassian: What if there are two?

ash2k: Two shapes with the same name returned by an autowiring functions is an error. https://github.com/atlassian/voyager/pull/78

pkg/execution/plugins/atlassian/iamrole/iam_role_manager.go
ash2k: Thank amazon. The type reflects JSON structure of iam policy object. `resource` can be either a string or an array in their policy. 🤷‍♂️

pkg/orchestration/wiring/ec2compute/common/common.go
brjacobsatlassian: From a single wiring function, sure. But what stops that from happening from two different wiring plugins?

PR #79, opened 2019-01-24
Add a bump-dependencies target to be used in a bamboo job

nilebox:


Makefile
```diff
@@ -82,6 +82,13 @@ verify-vendor:
 
 #===============================================================================
 
+.PHONY: bump-dependencies
+bump-dependencies:
+	dep ensure -update github.com/atlassian/smith
```
ash2k: I think it'd be useful to have `-v` here to see the progress, even in CI.

Makefile
```diff
@@ -82,6 +82,13 @@ verify-vendor:
...
+	make update-vendor
```
ash2k: I think it'd be better to not call make from make. You can move the line above into a separate target and then change `bump-dependencies` to depend on both targets.

brjacobsatlassian: Makefiles calling Makefiles are the path to insanity.

I worked at a place once where they used Makefiles to generate Makefiles. That might be the only thing worse than this.

PR #84, opened 2019-01-25
Upgrade Smith

ychen-atlassian:


ychen-atlassian:
Oh no https://github.com/atlassian/voyager/pull/82 exists already.

PR #86, opened 2019-01-25
NONE: Migrate to ops-api compatible reporting api schema

nickhiggs:
The reporting API response needs to be nested to be compatible with the ops api. This changes to support both the current and new responses so providers can migrate.


pkg/reporter/namespace_report.go
```diff
@@ -225,25 +229,31 @@ func (n *NamespaceReportHandler) getResourceFromProvider(ctx context.Context, pr
 		return handleProviderError(logger, obj, errors.Errorf("Failed request responded with status %d body %s", resp.StatusCode, string(body)), "request")
 	}
 
-	providerResp := ProviderResponse{}
-	err = json.Unmarshal(body, &providerResp)
-	if err != nil {
-		return handleProviderError(logger, obj, err, "unmarshal json")
+	providerOpsResp := ProviderOpsResponse{}
+	err = json.Unmarshal(body, &providerOpsResp)
+	if err != nil || providerOpsResp.Result == nil {
```
ash2k: I don't see how `err != nil` can be caused by `Provider using old report spec`? These are distinct cases and I think should be handled separately in code.

pkg/reporter/namespace_report.go
```diff
@@ -68,6 +68,10 @@ type NamespaceReportHandler struct {
 	location voyager.Location
 }
 
+type ProviderOpsResponse struct {
+	Result *ProviderResponse `json:"report,omitempty"`
```
ash2k: Field name and json tag do not match. They probably should, at least to follow the convention that we have.

nickhiggs:
> Not opposed to it, but I'd like to understand the reason for this change.

In the ops api specification we specify a response for a synchronous ops command should be nested in a 'result' field. The Reporting API should match this because it is just a special ops api so we need this to allow services to migrate from the old schema response to the new one.

PR #89, opened 2019-01-29
Ensure all admission responses include reason and code

ychen-atlassian:
So there's a couple of thing:

1. If you don't provide a Reason and Code, the admission webhook will spit out a 500, which is not what you want.
2. Bad Request is not what you want. Bad Request happens when the request itself is malformed, and Kubernetes handles this deserialization for you. Instead, we should be returning 422 + Invalid, which controllers/kubectl will handle as "the actual contents of your request sucks and cannot work"
3. Unauthorized implies missing headers, which Kubernetes handles too. We want to return Forbidden.


pkg/execution/svccatadmission/micros_test.go
```diff
@@ -47,19 +48,19 @@ func TestMicrosAdmitFunc(t *testing.T) {
...
 			buildAdmissionReview(elsieMicros2Service, k8s.ServiceInstanceGVR, admissionv1beta1.Create, buildV1ServiceInstance(t, elsieComputeService)),
-			buildAdmissionResponse(true, 0, nil, `service central owner of service "elsie-compute-service" (elsie) is same as micros2 service "elsie-micros2-service" (elsie)`),
+			buildAdmissionResponse(true, 0, metav1.StatusReasonUnknown, nil, `service central owner of service "elsie-compute-service" (elsie) is same as micros2 service "elsie-micros2-service" (elsie)`),
 			false,
 		},
 		{
 			"ServiceHasDifferentOwnerForbidden",
 			buildAdmissionReview(elsieMicros2Service, k8s.ServiceInstanceGVR, admissionv1beta1.Create, buildV1ServiceInstance(t, dougComputeService)),
-			buildAdmissionResponse(false, http.StatusUnauthorized, nil, `service central owner of service "doug-compute-service" (doug) is different to micros2 service "elsie-micros2-service" (elsie)`),
+			buildAdmissionResponse(false, http.StatusForbidden, metav1.StatusReasonForbidden, nil, `service central owner of service "doug-compute-service" (doug) is different to micros2 service "elsie-micros2-service" (elsie)`),
```
scottgreenup: TIL: https://stackoverflow.com/a/6937030

PR #90, opened 2019-01-29
DENG-2157 adding pact test voyager <-> trebuchet

qunayu:
more info on Jira: [DENG-2157](https://hello.atlassian.net/browse/DENG-2157)


pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,138 @@
+package releases
+
+import (
+	"fmt"
+	"github.com/atlassian/voyager/pkg/releases/deployinator/models"
```
ychen-atlassian: Put this import at the bottom with the other github imports

ychen-atlassian:
You need to merge master to fix the builds

pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,138 @@
...
+		Provider: "deployinator-trebuchet",
+		Host:     "localhost",
+		PactDir:  "generated/pacts",
+	}
+
+	defer pact.Teardown()
+
+	// Define the interactions in resolve and batchResolve separately, with the same pact
+
+	ResolveEndpointTest(pact)
```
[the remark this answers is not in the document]
qunayu: We wanted these two to generate a single pact contract for us (under the same pact struct), so if we split it into two, it would generate 2 seperate contracts.

pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,140 @@
...
+		Service:       "ServiceName",
+		ReleaseGroups: map[string]map[string]interface{}{},
+	}
+
+	// Define request with dummy variables and headers set
+	var resolveTest = func() error {
+		u := fmt.Sprintf("http://%s:%d/api/v1/resolve", pact.Host, pact.Server.Port)
+		req, err := http.NewRequest("GET", u,
+			strings.NewReader(
+				`{"service":"serviceName", "environment":"dev", "region":"us-east", "account": "1233"}`))
```
ash2k: Wait, are we sending this payload in a GET request?

pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+		if err != nil {
+			return err
+		}
+		if _, err = http.DefaultClient.Do(req); err != nil {
+			return err
+		}
+
+		req.Header.Set("Content-Type", "application/json")
```
ash2k: GET should not have a payload and hence does not need a `Content-Type`

ash2k: Also, headers should be set before the request is sent, not after =)

pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,140 @@
...
+		AddInteraction().
+		Given("There is a release group that matches the service and location given").
+		UponReceiving("A request to get the release group").
+		WithRequest(dsl.Request{
+			Method:  "GET",
+			Path:    dsl.String("/api/v1/resolve"),
+			Headers: dsl.MapMatcher{"Accept": dsl.String("application/json")},
+		}).
+		WillRespondWith(dsl.Response{
+			Status:  200,
```
ash2k: `http.StatusOK`

pkg/releases/consumer_test.go
qunayu: Oops

pkg/releases/consumer_test.go
qunayu: Nope, wanted to send them as parameters...let me use `req.URL.Query()` instead

pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,161 @@
...
+		u := fmt.Sprintf("http://%s:%d/api/v1/resolve", pact.Host, pact.Server.Port)
+		req, err := http.NewRequest("GET", u, nil)
+		q := req.URL.Query()
+		q.Add("service", "serviceName")
+		q.Add("environment", "dev")
+		q.Add("region", "us-east-1")
+		q.Add("account", "1233")
+		req.URL.RawQuery = q.Encode()
+
+		if err != nil {
```
ash2k: Error checking should be right after the place where an error might have happened: `req, err := http.NewRequest("GET", u, nil)`

pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,161 @@
...
+			},
+		}).
+		WillRespondWith(dsl.Response{
+			Status:  http.StatusOK,
+			Headers: dsl.MapMatcher{"Content-Type": dsl.String("application/json")},
+			Body:    singleReleaseGroup,
+		})
+
+	// Verify the interaction
+	if err := pact.Verify(resolveTest); err != nil {
```
ash2k: Could use `require.NoError(t, err)` like everywhere else.

qunayu: This is a function that returns an error variable though..maybe leaving it for pact to handle would be better?

ash2k: Maybe I was not clear... I'm suggesting to do this:
```go
err := pact.Verify(resolveTest)
require.NoError(t, err)
```
This is almost the same to what we have here atm, but consistent with the style of other code.

qunayu: Ah okay, sorry, I see. Thanks

pkg/releases/consumer_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/atlassian/voyager/pkg/releases/deployinator/models"
+	"github.com/pact-foundation/pact-go/dsl"
+)
+
+/**
+This test tests the consumer interactions surrounding the resolve endpoints between
+voyager (consumer) and trebuchet (provider).
+
+To validate the pact against the current provider specs locally, use the contract testing cli found here:
+https://hello.atlassian.net/wiki/spaces/TESTA/pages/128609304/Contract+Testing+Getting+Started+Guide+for+Consumers
```
ivorp: go/link it as per chat

qunayu:
Need to sort out dependencies before merging

PR #91, opened 2019-01-30
VYGR-355 check some names

ychen-atlassian:


pkg/orchestration/wiring/entangler.go
```diff
@@ -270,6 +271,22 @@ func (w *worker) entangle(resource *orch_v1.StateResource, stateMeta *meta_v1.Ob
...
+		Name:         resource.Name,
+		Type:         resource.Type,
+		WiringResult: *result,
+	}
+	w.allWiredResourcesList = append(w.allWiredResourcesList, result.Resources...)
+	return false, nil
+}
+
+func (w *worker) validateWireUp(resource *orch_v1.StateResource, result *wiringplugin.WiringResult) (bool, error) {
 	if w.allWiredResources[resource.Name] != nil {
```
ash2k: This check does not make sense to me - we've checked that the map does not have an entry for this name above (1rst line of this method). I know it was there before. We should remove it if it's not needed.

PR #94, opened 2019-01-31
Invalids are non retriable

ychen-atlassian:
This changes the behavior of our interactions with kubernetes.

Kubernetes on a failed CRD schema validation, (hopefully) admission webhooks, etc, will return a 422 + Invalid reason. This means your request is formed correctly but the contents are wrong. This is a non retriable request.

Then we also check BadRequest and a bunch of other things that really shouldn't happen, and aren't really retriable.


pkg/k8s/updater/BUILD.bazel
```diff
@@ -22,3 +22,23 @@ go_library(
         "//vendor/k8s.io/client-go/tools/cache:go_default_library",
     ],
 )
+
+go_test(
```
ash2k: size missing

pkg/k8s/updater/objectupdater_test.go
```diff
@@ -0,0 +1,379 @@
...
+		// Can not throw Conflict so not tested here
+	}
+
+	for name, c := range cases {
+		shouldBeConflict := c.ShouldBeConflict
+		shouldBeRetriable := c.ShouldBeRetriable
+		errorToThrow := c.Error
+
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
```
ash2k: We stopped doing parallel subtests a while ago.

pkg/k8s/updater/objectupdater_test.go
```diff
@@ -0,0 +1,379 @@
...
+		// Can not throw AlreadyExists so not tested here
+	}
+
+	for name, c := range cases {
+		shouldBeConflict := c.ShouldBeConflict
+		shouldBeRetriable := c.ShouldBeRetriable
+		errorToThrow := c.Error
+
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
```
ash2k: We stopped doing parallel subtests a while ago

PR #107, opened 2019-02-02
VYGR-391: Get team from service-central in sc.GetService

fcobb:
Adds a call in service-central client during GetService that attempts to find the OpsGenie team for a given service.


pkg/servicecentral/client.go
```diff
@@ -389,3 +444,21 @@ func convertV2ServiceToV1(v2Service V2Service) ServiceData {
 	service.CreationTimestamp = &creationTimestamp
 	return service
 }
+
+func findOpsGenieTeamServiceAttribute(attributes []serviceAttributeResponse) (s ServiceAttribute, found bool, err error) {
+	// https://developer.atlassian.com/platform/service-central/rest/#api-api-v2-schemas-attributes-get
```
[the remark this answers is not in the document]
fcobb: Yeah, to be fair this is an Atlassian specific REST client and whole repo is riddled with Atlassian internals. 
I'll remove this comment.

pkg/servicecentral/testdata/get_service_attributes.rsp.json
```diff
@@ -0,0 +1,62 @@
+[
```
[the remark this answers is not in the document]
jokeyrhyme: Elsewhere, to cope with data that may have been created during this bug, we assume 0 or more OpsGenie attributes, but we select the one that was modifedOn last
It's actually a shame they fixed the bug this way, as supporting multiple teams might have been interesting :)

pkg/servicecentral/client.go
```diff
@@ -389,3 +444,20 @@ func convertV2ServiceToV1(v2Service V2Service) ServiceData {
...
+		if attr.Schema.Name != opsGenieSchemaName {
+			continue
+		}
+
+		team, ok := attr.Value["team"]
+		if !ok {
+			return ServiceAttribute{}, false, errors.Errorf("expected to find team name within schema of name %q", opsGenieSchemaName)
+		}
+
+		return ServiceAttribute{Team: team}, true, nil
```
jokeyrhyme: It looks like this selects the first OpsGenie attribute, but I wonder if the DB doesn't still contain records that have multiple OpsGenie attributes? 🤷‍♂️

pkg/servicecentral/it/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+        "//pkg/util:go_default_library",
+        "//pkg/util/auth:go_default_library",
+        "//pkg/util/logz:go_default_library",
+        "//pkg/util/pkiutil:go_default_library",
+        "//pkg/util/testutil:go_default_library",
+        "//vendor/github.com/stretchr/testify/require:go_default_library",
+        "//vendor/k8s.io/api/core/v1:go_default_library",
+        "//vendor/k8s.io/client-go/kubernetes/scheme:go_default_library",
+    ],
+)
```
fcobb: I copied this from the internal repo, I have mostly no idea what I am doing here.

pkg/servicecentral/client.go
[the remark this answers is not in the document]
nilebox: > It exposes Atlassian internals which it shouldn't.

@roaanv to be fair, half of this repo exposes Atlassian internals, and it already has links to Hello.
@ash2k has discussed this recently, and his argument was that you can find links to Hello on the internet outside this repo too.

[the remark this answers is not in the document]
ash2k: Whatever we decide to do, let's be consistent. If we don't want them - remove them all. I think it is ok to replace them with `go/voyager_oss/bla` links.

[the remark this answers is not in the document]
scottgreenup: I also think that we should avoid as much internal stuff as possible to move towards a more open-source friendly repo. If half of our stuff points at internal stuff and internal documentation, it means it's more difficult to consume for the public..

pkg/servicecentral/client_test.go
```diff
@@ -265,7 +265,90 @@ func TestGetService(t *testing.T) {
...
 	// then
 	require.NoError(t, err)
-	require.Equal(t, 1, handler.RequestSnapshots.Calls())
+	require.Equal(t, 2, handler.RequestSnapshots.Calls())
+}
+
+func TestGetServiceWithOpsGenieAttribute(t *testing.T) {
+	t.Parallel()
+	// given
+	handler := MockHandler(Match(
```
scottgreenup: nit: This doesn't jive with Go standards, and this indenting looks a bit confusing. Could you clean this up?

fcobb: Yeah no worries, interesting that our linting build didn't pick this up.

pkg/servicecentral/client_test.go
```diff
@@ -265,7 +265,90 @@ func TestGetService(t *testing.T) {
 
 	// then
 	require.NoError(t, err)
-	require.Equal(t, 1, handler.RequestSnapshots.Calls())
+	require.Equal(t, 2, handler.RequestSnapshots.Calls())
+}
+
+func TestGetServiceWithOpsGenieAttribute(t *testing.T) {
```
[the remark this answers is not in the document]
ash2k: There is no need to put everything into 1 test.

pkg/servicecentral/it/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_test")
+
+go_test(
+    name = "go_default_test",
+    testonly = False,
```
ash2k:
```suggestion
    size = "medium",
```
Don't need `testonly` (in the other place it allows `pkg_tar` rule to depend on the e2e test binary so that it can package it into a tar) but need `size` (tells bazel how "big" is the test. Mostly to avoid warnings).

pkg/servicecentral/it/client_manual_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	testLogger.Sugar().Infof("Number of returned attributes: %v", len(resp))
+	testLogger.Sugar().Infof("Attributes: %#v", resp)
+
+	bytes, _ := json.Marshal(resp)
+	testLogger.Sugar().Infof("Attributes JSON: %#v", string(bytes))
+}
+
+// data should be "export CENTRAL_YAML=$(kubectl -n voyager get secrets asap-creator -o yaml)"
+func getSecret(t *testing.T) *v1.Secret {
```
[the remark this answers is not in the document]
ash2k: I think this test should be in the internal repo. All e2e and integration test for internal systems are there.

pkg/servicecentral/it/client_manual_test.go
```diff
@@ -0,0 +1,69 @@
...
+	c := servicecentral.NewServiceCentralClient(testLogger, client, asapConfig, parseURL(t, defaultServiceCentralURL))
+
+	// Get Service Attributes
+	resp, err := c.GetServiceAttributes(ctx, auth.ToOptionalUser(testUser), "slime")
+	require.NoError(t, err)
+
+	testLogger.Sugar().Infof("Number of returned attributes: %v", len(resp))
+	testLogger.Sugar().Infof("Attributes: %#v", resp)
+
+	bytes, _ := json.Marshal(resp)
```
ash2k: Check error.

pkg/servicecentral/it/client_manual_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"testing"
+
+	"github.com/atlassian/voyager/pkg/servicecentral"
+	"github.com/atlassian/voyager/pkg/util"
+	"github.com/atlassian/voyager/pkg/util/auth"
+	"github.com/atlassian/voyager/pkg/util/logz"
+	"github.com/atlassian/voyager/pkg/util/pkiutil"
+	"github.com/atlassian/voyager/pkg/util/testutil"
+	"github.com/stretchr/testify/require"
+	"k8s.io/api/core/v1"
```
ash2k:
```suggestion
	core_v1 "k8s.io/api/core/v1"
```

pkg/servicecentral/it/client_manual_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	defaultServiceCentralURL = "https://services.prod.atl-paas.net"
+	fakeTestUsername         = "fcobb"
+)
+
+// You can run this file from root using
+// `go test -v pkg/servicecentral/it/client_manual_test.go`
```
ash2k: "how to use go" should live somewhere else 😛 i.e. we don't have such instructions anywhere else, why here?

pkg/servicecentral/it/client_manual_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	client := util.HTTPClient()
+	c := servicecentral.NewServiceCentralClient(testLogger, client, asapConfig, parseURL(t, defaultServiceCentralURL))
+
+	// Get Service Attributes
+	resp, err := c.GetServiceAttributes(ctx, auth.ToOptionalUser(testUser), "slime")
+	require.NoError(t, err)
+
+	testLogger.Sugar().Infof("Number of returned attributes: %v", len(resp))
+	testLogger.Sugar().Infof("Attributes: %#v", resp)
```
ash2k: Use `t.Logf()` directly. And perhaps can actually just remove this logging? Can be added for debugging when needed.

ash2k: There should be some assertions on `resp`. Just printing the result is not very useful for automated testing.

pkg/servicecentral/types.go
```diff
@@ -30,11 +30,36 @@ type ServiceData struct {
 	Stateless            bool   `json:"stateless,omitempty"`
 	BusinessUnit         string `json:"business_unit,omitempty"`
 
+	Attributes []ServiceAttribute
```
ash2k: json tags are missing here and below.

pkg/servicecentral/types.go
```diff
@@ -30,11 +30,36 @@ type ServiceData struct {
...
 	Compliance *ServiceComplianceConf `json:"compliance,omitempty"`
 }
 
+type ServiceAttribute struct {
+	Team string
+}
+
+type ServiceAttributeResponse struct {
+	ID      int `json:"id"`
+	Service struct {
```
ash2k: Please don't use anonymous structs like that. Just to be consistent with the rest of the repo.

pkg/servicecentral/it/client_manual_test.go
fcobb: My understanding was that this is a manual test of the logic rather than an integration test. Does your comment still hold true in the case where its not an integration test?

pkg/servicecentral/client.go
```diff
@@ -294,7 +295,23 @@ func (c *Client) GetService(ctx context.Context, user auth.OptionalUser, service
...
-	return &parsedBody.Data[0], nil
+	service := &parsedBody.Data[0]
+
+	resp, err := c.GetServiceAttributes(ctx, user, serviceUUID)
+	if err != nil {
+		l.Error("Failed to get attributes for service", zap.Error(err), zap.String("service", serviceUUID))
+	}
+	ogTeamAttr, found, err := findOpsGenieTeamServiceAttribute(resp)
+	if err != nil {
+		// We do not return an error here as OpsGenie team is currently optional, likely to change when we remove PagerDuty
```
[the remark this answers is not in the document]
nilebox: Error to fetch attributes is different from missing attributes, so we shouldn't ignore the errors completely.
Fine to change the log level to Info/Warn though.

pkg/servicecentral/it/client_manual_test.go
ash2k: What exactly is manually tested? Why cannot those manual checks be encoded here? If you decide to keep logging, use `t` for it directly in tests, not zap.

pkg/servicecentral/client.go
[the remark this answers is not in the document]
fcobb: My thought was that as we only rely on service attributes for an optional attribute, we should not return an error. If we return here then we can fail processing the service. Is that what you want?

pkg/servicecentral/client.go
```diff
@@ -328,6 +345,44 @@ func (c *Client) DeleteService(ctx context.Context, user auth.User, serviceUUID
 	return nil
 }
 
+// GetServiceAttributes queries service central for the attributes of a given service. Can return an empty array if no attributes were found
+func (c *Client) GetServiceAttributes(ctx context.Context, user auth.OptionalUser, serviceUUID string) ([]ServiceAttributeResponse, error) {
+	req, err := c.rm.NewRequest(
+		pkiutil.AuthenticateWithASAP(c.asap, asapAudience, user.NameOrElse(noUser)),
+		restclient.Method(http.MethodGet),
+		restclient.JoinPath(fmt.Sprintf(v2ServicesPath+"/%s/attributes", url.PathEscape(serviceUUID))),
```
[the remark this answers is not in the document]
fcobb: I wrote a simple test to double check my understanding:

```
func TestFoo(t *testing.T) {
	serviceUUID := "slime-service"
	rm := restclient.NewRequestMutator(
		restclient.BaseURL("https://www.atlassian.com"),
	)

	req, err := rm.NewRequest(
		restclient.JoinPath(fmt.Sprintf(v2ServicesPath+"/%s/attributes", url.PathEscape(serviceUUID))),
	)
	if err != nil {
		t.Log(err)
	}
	t.Logf("URL: %q", req.URL)
}
```

--- PASS: TestFoo (0.00s)
    client_test.go:126: URL: "https://www.atlassian.com/api/v2/services/slime-service/attributes"
PASS

pkg/servicecentral/it/client_manual_test.go
fcobb: Heres where I took this idea from: https://github.com/atlassian/voyager/blob/master/pkg/pagerduty/it/client_manual_test.go#L28

pkg/servicecentral/types.go
```diff
@@ -30,11 +30,40 @@ type ServiceData struct {
...
 	// Compliance is a read-only field. It can be nil, in which case it means
 	// they have not completed their compliance questions yet
 	Compliance *ServiceComplianceConf `json:"compliance,omitempty"`
 }
 
+type ServiceAttribute struct {
+	Team string `json:"team,omitempty"`
+}
+
+type ServiceAttributeResponse struct {
```
fcobb: If/when we have more service attributes it is likely that we will need to change these types.

pkg/servicecentral/client.go
```diff
@@ -391,3 +444,28 @@ func convertV2ServiceToV1(v2Service V2Service) ServiceDataRead {
...
+		if !ok {
+			return ogTeamAttr, found, errors.Errorf("expected to find team name within schema of name %q", opsGenieSchemaName)
+		}
+
+		ogTeamAttr = ServiceAttribute{Team: team}
+		found = true
+		count++
+	}
+	if count > 1 {
+		return ogTeamAttr, found, errors.New("found more than one OpsGenie service attribute")
```
nilebox: Maybe return the number of found teams in the message as well?

PR #108, opened 2019-02-03
Auto-bump dependencies

atlassian-bamboo-agent:


ash2k:
@nilebox would be great to add "Red" to reviewers when such PR is opened.

p.s. It works! I've just merged a pr in smith and a minute later this one popped up.

PR #110, opened 2019-02-03
Cleanup status construction code

ash2k:


ash2k:
`formationobjectresult` is not part of any api so should not be exported? Both fields that it has are private already and it is not used (and should not be) in any other package.

ash2k:
Removing pointers because usually slices and maps contain values directly because that way allocation of the needed space is done in a single block vs many small allocations. `formationobjectresult` is a small object, it consists of 2 pointers only.

PR #113, opened 2019-02-04
SQS dead letter queue is only populated on MaxReceiveCount

ychen-atlassian:
As part of moving over our wiring to shapes I didn't take into consideration that the dead letter queue things only exist if you provide MaxReceiveCount.


pkg/orchestration/wiring/sqs/plugin.go
```diff
@@ -79,18 +79,31 @@ func WireUp(stateResource *orch_v1.StateResource, context *wiringplugin.WiringCo
 	}
 	wiredResources = append(wiredResources, serviceInstance)
 
+	var hasDeadLetterQueue bool
+	if stateResource.Spec != nil {
+		var spec map[string]interface{}
```
ash2k: I think it'd be better to unmarshal into an anonymous struct with just one field to avoid storing all the results that are not need.

pkg/orchestration/wiring/testdata/ec2compute.basicdep.bundle.yaml
```diff
@@ -63,22 +63,10 @@ spec:
           secretName: compute--basic
   - name: compute--secret
     references:
```
nilebox: do you want to add another queue to the state yaml, which _will_ have dead letter environment variables?

nilebox: or create a separate test state

ychen-atlassian: The E2E tests do test this already so it's good.

PR #116, opened 2019-02-05
VYGR-411: Use a simple cache for Service Central

nilebox:
This is a simple implementation of a cache which should reduce the number of requests sent to Service Central without changing the actual logic.

### DIFF
--- a/pkg/synchronization/controller.go
+++ b/pkg/synchronization/controller.go
@@ -458,7 +458,7 @@ func (c *Controller) createOrUpdateServiceMetadata(logger *zap.Logger, ns *core_
 	return false, nil
 }
 
-func (c *Controller) getCachedServiceData(name voyager.ServiceName) (*creator_v1.Service /* ok */, bool) {
+func (c *Controller) getCachedServiceData(name voyager.ServiceName) (*creator_v1.Service, bool /* ok */) {
 	serviceCacheMutex.RLock()
 	defer serviceCacheMutex.RUnlock()
 
--- a/pkg/synchronization/controller.go
+++ b/pkg/synchronization/controller.go
@@ -142,8 +142,12 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (bool /* retriable */, er
 	ctx.Logger.Sugar().Infof("Looking up service data for service %q from Service Central", serviceName)
 	serviceData, ok := c.getCachedServiceData(serviceName)
 	if !ok {
-		// no retries if SC is missing a service record
+		// no retries if service is missing in cache
 		return false, errors.New("service not found in cache")
+	}
+	if serviceData == serviceTombstone {
+		// no retries if service is missing in SC
+		return false, errors.New("service does not exist in Service Central")
 	}
 
 	operations := []func() (bool, error){
--- a/pkg/synchronization/controller.go
+++ b/pkg/synchronization/controller.go
@@ -69,6 +69,7 @@ const (
 var (
 	// in-memory cache of service records from Service Central
 	serviceCache map[voyager.ServiceName]*creator_v1.Service
+	serviceCacheMutex sync.Mutex
 )
 
 type ServiceMetadataStore interface {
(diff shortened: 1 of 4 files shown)


pkg/synchronization/controller.go
```diff
@@ -456,7 +461,25 @@ func (c *Controller) createOrUpdateServiceMetadata(logger *zap.Logger, ns *core_
 	return false, nil
 }
 
-func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.ServiceName) (*creator_v1.Service, error) {
+func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.ServiceName, allowCached bool) (*creator_v1.Service, error) {
```
nilebox: TODO: introduce a mutex to make sure there are no concurrent requests between a controller from the queue and a polling goroutine

nilebox: Actually I think we can only fetch data from the cache for namespace sync (the item is guaranteed to be present in cache, otherwise SD cannot be created without RBAC being setup by a polling loop) and use `sync.RWLock`.
Will rewrite a bit.

pkg/synchronization/controller.go
```diff
@@ -456,8 +455,31 @@ func (c *Controller) createOrUpdateServiceMetadata(logger *zap.Logger, ns *core_
...
+func (c *Controller) getCachedServiceData(name voyager.ServiceName) (*creator_v1.Service, bool /* ok */) {
+	c.ServiceCacheMutex.RLock()
+	defer c.ServiceCacheMutex.RUnlock()
+
+	service, ok := c.ServiceCache[name]
+	return service, ok
+}
+
+func (c *Controller) fetchAndCacheServiceData(name voyager.ServiceName) (*creator_v1.Service, error) {
+	c.ServiceCacheMutex.Lock()
```
ash2k: If I'm reading the code correctly this means that all the workers are serialized on this mutex and there is no concurrency here. We don't want that.

nilebox: I can move it after the call to SC, then it will be less impactful.

nilebox: to clarify: given that a single call to SC is the slowest part of this function, moving a lock after it means that the SC requests will be still parallelized, which is the main reason for executing this function from multiple go routines.

nilebox:
> I don't think local cache is the right approach here, we can do better than that. See my comment on the ticket.

@ash2k I've replied to your comment on that ticket. The main purpose of this PR is to address the issue with a long delay between adding item to the queue and processing it. I see 2 options of quickly fixing this issue:
1. Reduce the average processing time for each item - this PR achieves that via cache and reducing a processing of each item by 100-200 ms (time for a single call to SC)
2. Increase the number of workers, as a consequence of which we will increase the load on SC.

I'm not opposed to refactoring of the synchronization logic, I just don't want to spend too much time on that right now. If you're fine with increasing the number of workers instead of caching, I can work on that. The refactoring work could be part of the "technical debt" epic.

ash2k:
I'm just afraid that the needed refactoring will not happen with this band aid in place. I'd prefer a faster and simpler solution of increasing the number of workers. Don't delete this branch though =)

nilebox:
The fix increasing workers and QPS has been merged, and the monitor alert has been resolved.
So this is not urgent anymore, and we can work on a full refactoring of synchronization controller if we want :)

pkg/synchronization/controller.go
```diff
@@ -133,14 +136,10 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (bool /* retriable */, er
...
-		if servicecentral.IsNotFound(err) {
-			// no retries if SC is missing a service record
-			return false, err
-		}
-		// should be able to retry other SC errors
-		return true, err
+	serviceData, ok := c.getCachedServiceData(serviceName)
+	if !ok {
+		// no retries if service is missing in cache
+		return false, errors.New("service not found in cache")
```
ychen-atlassian: Should this not try to fetch if the service is not there? The only reason it's not there is on initial service create right?

nilebox: @ychen-atlassian actually, even on creation it shouldn't be possible to run into this, because user won't have permissions to create SD until we create a ClusterRoleBinding for SD, which requires fetching (and caching) data from Service Central.

So really this should never happen, hence no need to fetch.

nilebox:
This is not urgent anymore, and can be fixed with a proper refactoring. Closing.

PR #117, opened 2019-02-05
Cleanup some bits in Synchronization

ash2k:
Was looking at it for some unrelated reason and found some things that need to be cleaned up.


pkg/synchronization/controller.go
```diff
@@ -497,52 +496,58 @@ func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta
...
 	}
 
-	// Set the namespace in the spec to the one we intend to copy to
-	secret.Namespace = namespace.Name
-
-	// Reset the UID, owner references and resourceVersion
-	secret.UID = ""
-	secret.ObjectMeta.OwnerReferences = nil
-	secret.ResourceVersion = ""
+	secret := &core_v1.Secret{
```
ash2k: Instead of mutilating an existing object construct a new one, copying only the needed things.

pkg/synchronization/controller.go
```diff
@@ -497,52 +496,58 @@ func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta
...
-			return true, nil, err
+			return true, err
 		}
 		exists = false
 	}
 
 	if exists {
-		existingSecret.Type = secretSpec.Type
-		existingSecret.Data = secretSpec.Data
-		secret, err := c.MainClient.CoreV1().Secrets(secretSpec.Namespace).Update(existingSecret)
```
ash2k: Previously this code was always sending an update even if the data is the same?

PR #118, opened 2019-02-05
MICROSHELP-5236: [DO NOT MERGE] Check the service does not exist before creating

scottgreenup:
Currently we error out because of something like:
```
ERROR: failed to create service: Internal error occurred: Failed to update PagerDuty metadata in Service Central: BadRequest: failed to update service "sgreenup-test-1". Response: {
  "data": [],
  "message": "Cannot update owner/admins of service sgreenup-test-1 when SSAM container is set. Please update owner/admins using SSAM",
  "meta": null,
  "status_code": 400
}
ERROR: For assistance, try: atlas service --help
```

But this is not clear that the service already exists and that is why there is an error. I've added a check for an existing service.


scottgreenup:
This is flagged as "DO NOT MERGE" as I need to deploy it and see if it works :)

scottgreenup:
I was misinformed. We actually want this as a way of touching/updating the service metadata.

PR #120, opened 2019-02-05
VYGR-391: Add opsgenie integrations to notifications

fcobb:
**What:** This PR uses the Opsgenie integration manager client to add Opsgenie integrations into the service-metadata configmap, allowing autowiring functions to make use of the integrations.
**Opsgenie integration manager client PR:** https://github.com/atlassian/voyager/pull/119

**Manually tested the following scenarios in the playground:**
* Creating/deploying a service without opsgenie team then adding it, verifying that the configmap was updated.
* Services with invalid opsgenie team will return an error back through processing.
* Services without Opsgenie team will successfully process.


cmd/synchronization/app/controller_constructor.go
```diff
@@ -79,6 +80,10 @@ func (cc *ControllerConstructor) New(config *ctrl.Config, cctx *ctrl.Context) (*
 	scHTTPClient := util.HTTPClient()
 	scClient := servicecentral.NewServiceCentralClient(config.Logger, scHTTPClient, opts.ASAPClientConfig, opts.Providers.ServiceCentralURL)
 
+	// create a client for talking to OpsGenie Integration Manager
+	ogHTTPClient := util.HTTPClient()
```
ash2k: Why are we creating a some many http clients?

cmd/synchronization/app/options.go
```diff
@@ -43,7 +45,16 @@ func (p *Providers) UnmarshalJSON(data []byte) error {
 	}
 	depURL, err := url.Parse(rawProviders.Deployinator)
 	p.DeployinatorURL = depURL
-	return errors.Wrap(err, "unable to parse Deployinator URL")
+	if err != nil {
+		return errors.Wrap(err, "unable to parse Deployinator URL")
+	}
+
+	ogUrl, err := url.Parse(rawProviders.OpsgenieIntegrationsManager)
+	p.OpsgenieIntegrationsManagerURL = ogUrl
```
ash2k: Move below after error check

pkg/apis/creator/v1/types.go
```diff
@@ -100,6 +101,10 @@ type PagerDutyIntegrationMetadata struct {
 	IntegrationKey string `json:"integrationKey,omitempty"`
 }
 
+type OpsGenieMetadata struct {
+	Team string
```
ash2k: json tag missing

pkg/servicecentral/metadata.go
```diff
@@ -29,6 +32,21 @@ func SetPagerDutyMetadata(serviceCentralData *ServiceData, m *creator_v1.PagerDu
 	return setMetadata(serviceCentralData, PagerDutyMetadataKey, m)
 }
 
+// GetOpsGenieMetadata reads the opsgenie metadata out of a service
+func GetOpsGenieMetadata(serviceCentralData *ServiceData) (*creator_v1.OpsGenieMetadata, error) {
```
ash2k: Why are we doing procedural programming here in this file? :D

ash2k: Why these helper functions are public?

pkg/servicecentral/metadata.go
```diff
@@ -29,6 +32,21 @@ func SetPagerDutyMetadata(serviceCentralData *ServiceData, m *creator_v1.PagerDu
...
+		return nil, err
+	}
+	return &m, nil
+}
+
+// SetOpsGenieMetadata stores the metadata for opsgenie into a Service's metadata
+func SetOpsGenieMetadata(serviceCentralData *ServiceData, m *creator_v1.OpsGenieMetadata) error {
+	return setMetadata(serviceCentralData, OpsGenieMetadataKey, m)
+}
+
```
ash2k: `if m == nil {` below does not look right to me.

pkg/synchronization/controller.go
```diff
@@ -57,13 +58,15 @@ const (
...
-	baseDelayProcSec                    = 15
-	rmsPollJitterFactor                 = 1.2
+	serviceCentralPollPeriod                               = 30 * time.Second
+	serviceCentralListAllPeriod                            = 30 * time.Minute
+	serviceCentralListDriftCompensation                    = 5 * time.Second
+	releaseManagementPollPeriod                            = 5 * time.Second
+	releaseManagementSyncAllPeriod                         = 30 * time.Minute
+	baseDelayProcSec                                       = 15
+	rmsPollJitterFactor                                    = 1.2
+	secretTypeOpsGenie                  core_v1.SecretType = voyager.Domain + "/opsgenie"
```
ash2k: Does this actually work? I'm not sure kube will like custom secret types.

pkg/synchronization/controller.go
```diff
@@ -57,13 +58,15 @@ const (
...
+	secretNameOpsGenie                                     = "OpsGenieIntegrations"
```
ash2k: I think we use lowercase letters with dashes for names.

pkg/synchronization/controller.go
```diff
@@ -72,6 +75,10 @@ type ServiceMetadataStore interface {
 	ListModifiedServices(ctx context.Context, user auth.OptionalUser, modifiedSince time.Time) ([]creator_v1.Service, error)
 }
 
+type OpsGenieIntegrationManagerClient interface {
+	GetOrCreateIntegrations(ctx context.Context, teamName string) (bool, *opsgenieIntegrationManager.IntegrationsResponse, error)
```
ash2k: please swap first and second return values

pkg/synchronization/controller.go
```diff
@@ -436,6 +444,22 @@ func (c *Controller) createOrUpdateServiceMetadata(logger *zap.Logger, ns *core_
 		},
 	}
 
+	// OpsGenie team is currently optional
+	if serviceData.Spec.Metadata.OpsGenie != nil {
+		retriable, resp, err := c.OpsGenie.GetOrCreateIntegrations(context.Background(), serviceData.Spec.Metadata.OpsGenie.Team)
```
ash2k: Use `context.TODO()` when there is no context to pass.

pkg/synchronization/controller.go
```diff
@@ -452,6 +476,33 @@ func (c *Controller) createOrUpdateServiceMetadata(logger *zap.Logger, ns *core_
 	return false, nil
 }
 
+// createOpsGenieSecretSpec creates an OpsGenieIntegrations secret object suitable for Secret creation
+func createOpsGenieSecretSpec(ogInt *opsgenieIntegrationManager.IntegrationsResponse, ns *core_v1.Namespace) (core_v1.Secret, error) {
+	jData, err := json.Marshal(ogInt)
+	if err != nil {
+		return core_v1.Secret{}, err
```
ash2k: Use pointers for API types.

pkg/synchronization/controller.go
```diff
@@ -452,6 +476,33 @@ func (c *Controller) createOrUpdateServiceMetadata(logger *zap.Logger, ns *core_
...
+	}
+
+	secretData := map[string][]byte{
```
ash2k: Why not inline it into the struct literal?

pkg/synchronization/controller.go
```diff
@@ -452,6 +476,33 @@ func (c *Controller) createOrUpdateServiceMetadata(logger *zap.Logger, ns *core_
 	return false, nil
 }
 
+// createOpsGenieSecretSpec creates an OpsGenieIntegrations secret object suitable for Secret creation
+func createOpsGenieSecretSpec(ogInt *opsgenieIntegrationManager.IntegrationsResponse, ns *core_v1.Namespace) (core_v1.Secret, error) {
```
ash2k: No need to pass the whole namespace object, just the name is enough.

ash2k (dismissed):

pkg/servicecentral/metadata.go
```diff
@@ -10,6 +10,9 @@ const (
 	// PagerDutyMetadataKey is the key we use to store the Metadata in the Misc field
 	PagerDutyMetadataKey = "pagerduty"
 
+	// OpsGenieMetadataKey is the key we use to store the Metadata in the Misc field
+	OpsGenieMetadataKey = "opsgenie"
```
nilebox: We don't need this? We agreed not to store Opsgenie metadata in the Misc field.

pkg/apis/creator/v1/types.go
```diff
@@ -59,6 +59,7 @@ func (ss *ServiceSpec) EmailAddress() string {
 // +k8s:deepcopy-gen=true
 type ServiceMetadata struct {
 	PagerDuty *PagerDutyMetadata `json:"pagerDuty,omitempty"`
+	OpsGenie  *OpsGenieMetadata  `json:"opsGenie,omitempty"`
```
nilebox: This change seems valid, we should just set the team name based on the "team" service attribute rather than misc field. Also, this part should be read-only, we never edit team, editing is done via Microscope UI (at least that's our agreement for now)

pkg/servicecentral/metadata.go
```diff
@@ -29,6 +32,21 @@ func SetPagerDutyMetadata(serviceCentralData *ServiceData, m *creator_v1.PagerDu
 	return setMetadata(serviceCentralData, PagerDutyMetadataKey, m)
 }
 
+// GetOpsGenieMetadata reads the opsgenie metadata out of a service
+func GetOpsGenieMetadata(serviceCentralData *ServiceData) (*creator_v1.OpsGenieMetadata, error) {
+	var m creator_v1.OpsGenieMetadata
+	found, err := unmarshalFromMiscData(serviceCentralData, OpsGenieMetadataKey, &m)
```
nilebox: This is wrong, we should pull team name from service attribute and construct OpsGenieMetadata struct on the fly (no unmarshalling).

pkg/servicecentral/store.go
```diff
@@ -241,6 +241,13 @@ func serviceToServiceData(existingData ServiceData, service *creator_v1.Service)
 		}
 	}
 
+	if service.Spec.Metadata.OpsGenie != nil {
```
nilebox: This is a write path, we don't need this - team is readonly for us.

pkg/synchronization/controller.go
nilebox: I think we use "opaque" type everywhere (default), but if it works - why not

pkg/synchronization/controller.go
nilebox: also add comment for what bool means please

pkg/synchronization/controller.go
nilebox: I think we should define a separate struct in the `orchestration` package and convert from `opsgenieIntegrationManager.IntegrationsResponse` to it somewhere, similar to how we have defined `ServiceProperties` for `service-metadata` ConfigMap to be passed to autowiring functions. We don't want to couple autowiring context to Opsgenie Integration manager response format.

nilebox: Also I think we agreed to store a team name in one of the keys. This may be useful to see which team the integrations are associated with (in case the team changes in Microscope)

cmd/synchronization/app/controller_constructor.go
fcobb: 🤷‍♂️ Should I create one client and re-use for service central and opsgenie?

cmd/synchronization/app/options.go
fcobb: I copied the lines above it, would you like me to change those within this PR?

pkg/synchronization/controller.go
fcobb: Just testing you ;)

pkg/apis/creator/v1/types.go
fcobb: I don't know what you mean, lets talk IRL.

pkg/synchronization/controller.go
fcobb: This is not yet tested. 🤷‍♂️ Should I change it to "opaque"?

pkg/synchronization/controller.go
```diff
@@ -459,41 +465,65 @@ func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.Service
 	return c.ServiceCentral.GetService(context.Background(), user, servicecentral.ServiceName(name))
 }
 
-func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta.Notifications, error) {
+func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta.Notifications, bool /* retriable */, error) {
+	// Default pagerduty values re-used from Micros config.js
```
fcobb: Instead of return this at the end we now instantiate the notifications with defaults and override it if the pagerduty or opsgenie notifications creation is successful.

pkg/synchronization/controller.go
```diff
@@ -459,41 +465,65 @@ func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.Service
...
 			LowPriorityPagerdutyEndpoint: *lowPriPD,
-		}, nil
+		}
+	}
+	integrations, retriable, err := c.getOpsgenieIntegrations(spec.Metadata.Opsgenie)
+	if err != nil {
+		return nil, retriable, errors.Wrap(err, "failed to create opsgenie notifications")
+	}
+	ogNotifications, err := buildOpsgenieNotifications(integrations, c.ClusterLocation.EnvType)
+	if err == nil {
```
fcobb: This `err == nil` feels weird to me as I usually l like to keep the happy path on the first tab. I might have another go at reworking this function, let me know if you have any ideas.

nilebox: we should probably still log the error?
```
if err != nil {
  // log
} else {
  notifications.OpsgenieIntegrations = ogNotifications
}
```

pkg/synchronization/controller.go
```diff
@@ -459,41 +465,65 @@ func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.Service
...
 
-		return &orch_meta.Notifications{
+		notifications = orch_meta.Notifications{
 			Email:                        spec.EmailAddress(),
 			PagerdutyEndpoint:            *mainPD,
 			LowPriorityPagerdutyEndpoint: *lowPriPD,
-		}, nil
+		}
+	}
+	integrations, retriable, err := c.getOpsgenieIntegrations(spec.Metadata.Opsgenie)
```
nilebox: what happens if team is not specified in Service Central? Will we get an empty list of integrations?

fcobb: I do a check at the start of the called function to see if `spec.Metadata.Opsgenie` is nil
`func (c *Controller) getOpsgenieIntegrations(metadata *creator_v1.OpsgenieMetadata) ([]opsgenie.Integration, bool /* retriable */, error) {
	// Opsgenie is optional
	if metadata == nil {
		return nil, true, nil
	}`

pkg/synchronization/controller.go
```diff
@@ -459,41 +465,65 @@ func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.Service
 	return c.ServiceCentral.GetService(context.Background(), user, servicecentral.ServiceName(name))
 }
 
-func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta.Notifications, error) {
+func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta.Notifications, bool /* retriable */, error) {
+	// Default pagerduty values re-used from Micros config.js
+	notifications := orch_meta.Notifications{
+		Email: spec.EmailAddress(),
+		PagerdutyEndpoint: orch_meta.PagerDuty{
```
nilebox: we may define vars at the global level...
```
var
(
  defaultPagerdutyEndpoint = ...
  defaultLowPriorityPagerdutyEndpoint = ...
)
```

pkg/synchronization/controller.go
```diff
@@ -459,41 +465,65 @@ func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.Service
...
+			return nil, false, errors.Wrap(err, "cannot convert main pagerduty entry")
 		}
 		lowPriPD, err := convertPagerDuty(pdEnvMetadata.LowPriority)
 		if err != nil {
-			return nil, errors.Wrap(err, "cannot convert low priority pagerduty entry")
+			return nil, false, errors.Wrap(err, "cannot convert low priority pagerduty entry")
 		}
 
-		return &orch_meta.Notifications{
+		notifications = orch_meta.Notifications{
```
nilebox: Email remains the same, we could set `notifications.PagerdutyEndpoint` and `notifications.LowPriorityPagerdutyEndpoint` to make it more explicit which parts do we want to modify.

cmd/synchronization/app/controller_constructor.go
fcobb: As discussed outside this PR we will address this later.

cmd/synchronization/app/options.go
```diff
@@ -43,7 +45,17 @@ func (p *Providers) UnmarshalJSON(data []byte) error {
 	}
 	depURL, err := url.Parse(rawProviders.Deployinator)
 	p.DeployinatorURL = depURL
-	return errors.Wrap(err, "unable to parse Deployinator URL")
+	if err != nil {
+		return errors.Wrap(err, "unable to parse Deployinator URL")
+	}
```
fcobb: I am not sure why this always returned an error and I cannot identify where this function is called.

pkg/opsgenie/types.go
```diff
@@ -5,13 +5,22 @@ type IntegrationsResponse struct {
...
+	EnvType  EnvType `json:"envType"`
 }
+
+type EnvType string
+
+const (
+	EnvTypeDev     EnvType = "dev"
+	EnvTypeStaging EnvType = "staging"
+	EnvTypeProd    EnvType = "prod"
+	EnvTypeGlobal  EnvType = "null" // Intentionally a string called "null" as this is the expected result from opsgenie int manager
```
fcobb: cc @jokeyrhyme

pkg/servicecentral/client.go
```diff
@@ -300,14 +300,8 @@ func (c *Client) GetService(ctx context.Context, user auth.OptionalUser, service
...
 		return nil, errors.Wrap(err, "failed to get attributes for service")
 	}
-	ogTeamAttr, found, err := findOpsGenieTeamServiceAttribute(resp)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get OpsGenie attributes for service")
-	}
 
-	if found {
-		service.Attributes = append(service.Attributes, ogTeamAttr)
-	}
```
fcobb: Previously, we filtered the service attributes before adding them to the service object. We now do that filtering later on.

pkg/servicecentral/client_test.go
```diff
@@ -393,34 +394,6 @@ func TestGetServiceWithFailedAttributesCall(t *testing.T) {
 	require.Equal(t, 2, handler.RequestSnapshots.Calls())
 }
 
-func TestGetServiceWithMultipleOpsGenieAttribute(t *testing.T) {
```
fcobb: This test does not make sense anymore as it was making sure we errored on multiple OG attributes, this is no longer an error case as we handle multiple attributes later on.

pkg/synchronization/controller.go
```diff
@@ -54,7 +55,9 @@ const (
 	// see https://github.com/jtblin/kube2iam#namespace-restrictions
 	allowedRolesAnnotation = "iam.amazonaws.com/allowed-roles"
 
-	maxSyncWorkers = 10
+	maxSyncWorkers             = 10
+	defaultPagerdutyGeneric    = "5d11612f25b840faaf77422edeff9c76"
+	defaultPagerdutyCloudwatch = "https://events.pagerduty.com/adapter/cloudwatch_sns/v1/124e0f010f214a9b9f30b768e7b18e69"
```
fcobb: These defaults were in the `buildNotifcations` function.

pkg/synchronization/controller.go
```diff
@@ -459,41 +467,73 @@ func (c *Controller) getServiceData(user auth.OptionalUser, name voyager.Service
 	return c.ServiceCentral.GetService(context.Background(), user, servicecentral.ServiceName(name))
 }
 
-func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta.Notifications, error) {
+func (c *Controller) buildNotifications(spec creator_v1.ServiceSpec) (*orch_meta.Notifications, bool /* retriable */, error) {
```
fcobb: It is easier to understand these changes if you do a side-by-side diff.

cmd/synchronization/app/options.go
nilebox: seems like a bug, not sure how did it work

nilebox: it should be implicitly called in https://github.com/atlassian/voyager/blob/master/cmd/synchronization/app/options.go#L67

pkg/servicecentral/store.go
```diff
@@ -306,6 +306,11 @@ func serviceDataToService(data *ServiceDataRead) (*creator_v1.Service, error) {
 		service.Spec.Metadata.PagerDuty = pagerDutyMetadata
 	}
 
+	ogMetadata, _ := GetOpsGenieAttribute(data)
```
nilebox: shall we log an error?

pkg/synchronization/controller.go
```diff
@@ -665,6 +705,40 @@ func pagerDutyForEnvType(pagerduty *creator_v1.PagerDutyMetadata, envType voyage
...
+		switch envType {
+		case voyager.EnvTypeStaging:
+			if integration.EnvType == opsgenie.EnvTypeStaging {
+				filtered = append(filtered, integration)
+			}
+		case voyager.EnvTypeProduction:
+			if integration.EnvType == opsgenie.EnvTypeProd {
+				filtered = append(filtered, integration)
+			}
+		default:
```
nilebox: I would prefer to have `case voyager.EnvTypeDev` here and throw an error in the `default:` section flagging an unexpected/unknown envType.

pkg/servicecentral/store.go
fcobb: > shall we log an error?
Considering this is optional, Should this be an error or info message?

nilebox: I would say that failure to make a call to Opsgenie integration manager is still an error, so I would log it at the error level. We should react to being unable to fetch Opsgenie integrations rather than completely ignoring them.

@ash2k WDYT?

fcobb: This is not the path of calling OG int manager,  this call will attempt to find any opsgenie attributes within the service object.

We call opsgenie int manager during the `createOrUpdateServiceMetadata` function when building the notifications.

nilebox: > attempt to find any opsgenie attributes within the service object

If attribute not found, there won't be an error, but just a `nil` result AFAICT?
So then the error means that either the attribute format is invalid, or there is some bug in Voyager code, and we should also log with error, I think.

fcobb: Yep you are correct, an error here is when:

1. we found an issue with an Opsgenie attribute
2. we found more than one Opsgenie attribute.

cmd/synchronization/app/options.go
```diff
@@ -43,7 +45,17 @@ func (p *Providers) UnmarshalJSON(data []byte) error {
 	}
 	depURL, err := url.Parse(rawProviders.Deployinator)
 	p.DeployinatorURL = depURL
-	return errors.Wrap(err, "unable to parse Deployinator URL")
+	if err != nil {
```
ash2k: Error check should be before the `depURL` is used. OR do this instead:
```suggestion
        p.DeployinatorURL, err = url.Parse(rawProviders.Deployinator)
	if err != nil {
```
And same above for sc url (and below)

pkg/opsgenie/types.go
ash2k: ?

pkg/servicecentral/client.go
```diff
@@ -344,7 +338,8 @@ func (c *Client) DeleteService(ctx context.Context, user auth.User, serviceUUID
 }
 
 // GetServiceAttributes queries service central for the attributes of a given service. Can return an empty array if no attributes were found
-func (c *Client) GetServiceAttributes(ctx context.Context, user auth.OptionalUser, serviceUUID string) ([]ServiceAttributeResponse, error) {
+func (c *Client) GetServiceAttributes(ctx context.Context, user auth.OptionalUser, serviceUUID string) ([]ServiceAttribute, error) {
+	var svcAttrResp []ServiceAttribute
```
ash2k: What is the reason for this change? Returning `nil` is the idiomatic thing to do - it is immediately clear that the value is not supposed to have anything in it and should not be used. With this change it is ambiguous - maybe the slice is empty or maybe not? You are forced to read the code to see it's empty.

pkg/servicecentral/metadata.go
```diff
@@ -29,6 +29,16 @@ func SetPagerDutyMetadata(serviceCentralData *ServiceDataWrite, m *creator_v1.Pa
 	return setMetadata(serviceCentralData, PagerDutyMetadataKey, m)
 }
 
+// GetOpsGenieAttribute reads the Opsgenie team attribute out of a service
+func GetOpsGenieAttribute(serviceCentralData *ServiceDataRead) (*creator_v1.OpsgenieMetadata, error) {
```
ash2k: Capital `G` =)

ash2k: I wonder why do we have all these functions exported here?

pkg/synchronization/controller.go
```diff
@@ -665,6 +705,42 @@ func pagerDutyForEnvType(pagerduty *creator_v1.PagerDutyMetadata, envType voyage
 	}
 }
 
+// filterOpsgenieIntegrationsByEnv filters a given list of integrations by the EnvType
+func filterOpsgenieIntegrationsByEnv(integrations []opsgenie.Integration, envType voyager.EnvType) ([]opsgenie.Integration, error) {
+	if len(integrations) == 0 {
+		return nil, nil // Not an error as Opsgenie is optional
+	}
+
+	filtered := make([]opsgenie.Integration, 0, 4)
```
ash2k: We usually do not pre-allocate if we don't exactly know the number of items. Just a style thing.

pkg/synchronization/controller.go
ash2k: These values should come from configuration, not be constants. Please move them.

pkg/opsgenie/types.go
ash2k: Why not `global`?

pkg/opsgenie/it/client_manual_test.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	opsGenieIntManURL = "https://micros.prod.atl-paas.net"
+	opsgenieIntManURL = "https://micros.prod.atl-paas.net"
```
ash2k: Don't use real data in tests please. Use fake data =)

cmd/synchronization/app/options.go
```diff
@@ -21,15 +21,17 @@ type Options struct {
...
 }
 
 // UnmarshalJSON unmarshals our untyped config file into a typed struct including URLs
 func (p *Providers) UnmarshalJSON(data []byte) error {
 	var rawProviders struct {
-		ServiceCentral string `json:"serviceCentral"`
-		Deployinator   string `json:"deployinator"`
+		ServiceCentral              string `json:"serviceCentral"`
+		Deployinator                string `json:"deployinator"`
+		OpsgenieIntegrationsManager string `json:"opsgenieIntegrationsManager"`
```
ash2k: Would be much more readable to name both json fields and struct fields like they are named in the `Providers` struct (camel case for JSON). I don't remember if we use `URL` or `Url` for Json fields, check other places.

ash2k (changes requested):

pkg/opsgenie/types.go
fcobb: I am not sure why its not called "global. @jokeyrhyme Do you know why?

pkg/synchronization/controller.go
fcobb: I'll do this in a follow up PR as this PR is big enough.

pkg/servicecentral/metadata.go
fcobb: Good question, looks like they are all used within the same package. I'll fix these in a follow up PR.

pkg/synchronization/controller.go
fcobb: Happy to change this to make our style if you would like. We expect there to be exactly 4 but it is possible that they add more integrations in the future. What would you like to see in this scenario?

pkg/opsgenie/it/client_manual_test.go
fcobb: This is a manual integration test, I don't think fake data is applicable.

cmd/synchronization/app/options.go
fcobb: I'll add that to the list of follow up PRs, unless you want to block this PR on it.

ash2k (dismissed):
I think this is fine to be merged. Please address the issues/nits in a follow up PR. I think the manual opsgenie integration test should be moved to the internal repo.

ash2k (dismissed):

pkg/opsgenie/types.go
jokeyrhyme: @fcobb Well, it's not that things are global, it's just that they aren't explicitly environment-specific
Maybe this should actually mean Global, and maybe we should actually emit `"global"` instead of `"null"`
When we put this together, the most important thing was making sure our CloudWatch alarms (which are environment-specific) were covered, everything else was secondary
I'm wondering whether we should call this Null or something so that we can meaningfully transition to Global in future?

fcobb: @jokeyrhyme I went for Global as if its not environment specific then its essentially global.
As you own this API I will defer these decisions to you and your team. My preference is to return a specific value for these non-environment specific values rather than null.

nilebox:
As discussed, please add a test for Opsgenie integrations present in `ServiceProperties` (in `entangler_test.go`

fcobb:
> As discussed, please add a test for Opsgenie integrations present in `ServiceProperties` (in `entangler_test.go`

Done

nilebox (approved):

nilebox:
Make sure you block the PR in stash updating dependencies after you merge this change, and fix config.

PR #121, opened 2019-02-05
Auto-bump dependencies

atlassian-bamboo-agent:


ash2k:
@nilebox so this also needs to update `ctrl`, not just `smith` dependency.

nilebox:
> so this also needs to update `ctrl`, not just `smith` dependency.

@ash2k I thought that would happen automagically based on the version of `ctrl` used by `smith`? no?

ash2k:
@nilebox no as long as we explicitly ask dep to update a single dependency only.

PR #122, opened 2019-02-05
MICROS-6935 rename platformdns autowiring

marciogh:
no logic change, just package/error messages/unit tests changes


pkg/orchestration/wiring/platformdns/api/types.go
```diff
@@ -1,12 +1,12 @@
...
+package apiplatformdns
 
 import (
 	"github.com/atlassian/voyager"
 	"github.com/atlassian/voyager/pkg/servicecatalog"
 )
 
 const (
-	ResourceType                  voyager.ResourceType           = "InternalDNS"
+	ResourceType                  voyager.ResourceType           = "PlatformDNS"
```
nilebox: This is a breaking change. If we have any existing SDs with `InternalDNS` resource, they will break.
You may keep both the old and new resource type name for transition period until you migrate all SDs to a new name.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -9,7 +9,7 @@ import (
```
nilebox: rename the file to `platformdns.go`?

pkg/orchestration/wiring/platformdns/api/types.go
nilebox: Apparently there is no documentation for InternalDNS on DAC https://developer.atlassian.com/platform/micros2/reference/resources/
there should be one?

marciogh:
> What is the reason for this change? `InternalDNS` means something while `PlatformDNS` does not.

@ash2k `PlatformDNS` means `atl-paas.net` domain, which is not necessarily internal, it can be resolved from the outside network. @roaanv can provide more context here, as this change has been proposed by him, and it's related to the incoming Global Edge integration (non `atl-paas.net` domains)

pkg/execution/svccatadmission/internaldns.go
marciogh: I'll deal with webhooks reanming in a separate PR

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -55,7 +55,7 @@ func InternalDNSAdmitFunc(ctx context.Context, microsServerClient microsServerCl
 				},
 			}, nil
 		}
-		var internalDNSSpec apiinternaldns.Spec
+		var internalDNSSpec apiplatformdns.Spec
```
roaanv: This should probably be:
var platformDNSSpec apiplatformdns.Spec

pkg/orchestration/wiring/platformdns/api/types.go
mcoot: Yes, this is intentionally a breaking change made now to avoid needing to care about compatibility, had we rolled out to users first. There are no resources outside the Micros teams (mostly personal test services, one belonging to a DE service in dev).

pkg/execution/svccatadmission/internaldns_test.go
```diff
@@ -44,30 +44,30 @@ func TestInternalDNSAdmitFunc(t *testing.T) {
...
 			"internaldns multiple new.domain",
 			buildAdmissionReview(dougComputeService, k8s.ServiceInstanceGVR, admissionv1beta1.Create, buildServiceInstance(
-				t, apiinternaldns.ClusterServiceClassExternalID, apiinternaldns.ClusterServicePlanExternalID, apiinternaldns.Spec{
-					Aliases: []apiinternaldns.Alias{
+				t, apiplatformdns.ClusterServiceClassExternalID, apiplatformdns.ClusterServicePlanExternalID, apiplatformdns.Spec{
+					Aliases: []apiplatformdns.Alias{
 						{
-							AliasType: apiinternaldns.AliasTypeSimple,
+							AliasType: apiplatformdns.AliasTypeSimple,
 							Name:      "new1.domain",
```
roaanv: Not part of this PR, but I'm wondering if these test domains shouldn't be restricted to
new1.atl-paas.net
Since I expect there will be various rules later that check that the alias is a subdomain of atl-paas.net.
Or we can keep it until the tests break

marciogh: I like the idea of waiting the tests to break.

pkg/execution/svccatadmission/internaldns.go
marciogh: These files belong to the webhook, I have a separate ticket/PR for them

pkg/execution/svccatadmission/internaldns_test.go
mcoot: Right now, behaviour will be a failure on the provider side since a non-atl-paas domain will not be in the list of valid zones during provisioning.

marciogh: nit: not in provider side, on smith service catalog schema validation (before it actually makes the call to the provider)

PR #123, opened 2019-02-05
Strict YAML parsing

ash2k:
This allows us to detect more issues with invalid input. Everywhere we parse YAML we should use the `sigs.k8s.io/yaml` package's `UnmarshalStrict()`. This fork will also get bugfixes, unlike the original repo where nothing has been happening for months. Kube devs forked it for a reason.

As you can see we've had quite a bit of cruft in our fixtures. I wonder if we should use this mode for JSON parsing too. Not sure if it will work and if it will catch any issues. Needs testing.


pkg/execution/computionadmission/admission.go
```diff
@@ -219,7 +219,7 @@ func (ac *AdmissionContext) checkPRGBRequired(ar *admission_v1beta1.AdmissionReq
 
 	data := cfgMap.Data[orch_meta.ConfigMapConfigKey]
 	var config orch_meta.ServiceProperties
-	err = yaml.Unmarshal([]byte(data), &config)
+	err = yaml.UnmarshalStrict([]byte(data), &config)
```
nilebox: this maybe unnecessary? what if we add/remove some field in `service-metadata`? this may break a webhook (depending on deployment order)

pkg/formation/controller.go
```diff
@@ -131,7 +131,7 @@ func (c *Controller) processLocationDescriptor(logger *zap.Logger, ld *form_v1.L
 		if !ok {
 			return false, false, nil, errors.Errorf("release config map is missing expected data key: '%s'", releaseConfigMapDataKey)
 		}
-		err = yaml.Unmarshal([]byte(releaseStr), &releaseData)
+		err = yaml.UnmarshalStrict([]byte(releaseStr), &releaseData)
```
nilebox: what if there is a new field introduced on Deployinator side, but the client is not re-generated yet?

nilebox: ah, it's just `map[string]interface{}` rather than a struct, should be fine

pkg/execution/computionadmission/admission.go
ash2k: You are right, I haven't verified each usage place. I'll do that.

mmatusiak-atlassian:
Do we need a test somewhere that reads a deliberately invalid yaml file and asserts on the error message?

ash2k:
@mmatusiak-atlassian I don't think we need to test this. I think we need to test this in CLI and that's probably about it. We accept JSON in all API calls, not YAML. For JSON something else needs to be done separately (probably more webhooks).

cmd/orchestration/app/options.go
```diff
@@ -41,7 +41,7 @@ func readAndValidateOptions(configFile string) (*Options, error) {
 	}
 
 	var opts Options
-	if err := yaml.Unmarshal(doc, &opts); err != nil {
+	if err := yaml.UnmarshalStrict(doc, &opts); err != nil {
```
Burr: Belated question about this change - during our deployment process, do we roll back configmap changes if the deployment fails? If not, wouldn't this change potentially cause problem during failed deployments? Scenario being that we accidentally roll out bad config - at that point, not only will new versions of the app not start, but if any pods get rescheduled, they also won't start. We'd need to consciously roll back to the previous config/configmap if that doesn't happen automatically

PR #125, opened 2019-02-06
NONE: Add more information into error message about invalid location

nilebox:


pkg/replication/admission.go
```diff
@@ -275,7 +275,7 @@ func (ac *AdmissionContext) validateLocationsAndTransforms(sd *comp_v1.ServiceDe
 		clusterLocations.Insert(location)
 		if sdLocation.EnvType == ac.CurrentLocation.EnvType && !ac.ReplicatedLocations.Has(location) {
 			rejectionMessages = append(rejectionMessages,
-				fmt.Sprintf("location %q does not exist in %q", sdLocation.Name, sdLocation.EnvType))
+				fmt.Sprintf("location %q (region: %q, account: %q) does not exist in %q environment, see go/micros2-locations", sdLocation.Name, sdLocation.Region, sdLocation.Account, sdLocation.EnvType))
```
ash2k: This message is not OSS-friendly =)

cstivers78: The issue I had was I did not define an "account" in the SD. So, if account is required, it should say upfront "location dev does not specify an account"

nilebox: @ash2k I know, and happy to change the message once someone outside Atlassian complains about it :) for now I think it will be useful

nilebox: @cstivers78 good point, will check that as well.

nilebox: added more validation

ash2k:
Why not instead fix the schema here https://github.com/atlassian/voyager/blob/a64d9789a714ddc4a3d91bda9bb0bc8b8723205d/pkg/composition/crd/crd.go#L13-L38 ? =) Set min length and make account a required field. Also regex for label. And enum for envtype. And enum for region (list all aws regions).

nilebox:
> Why not instead fix the schema here

The error message will then be something like `"Invalid JSON: locations[2].account is a required field"`. The error message proposed in this PR is much more user friendly IMO.

nilebox:
I think this PR is actually a part of "SD validation UX" PC ticket
@kopper which solution would you prefer?

ash2k:
Are you suggesting we remove the schema and do all the validation by hand? =) I think schema should be used as much as possible. I think `Invalid JSON: locations[2].account is a required field` is pretty readable. Our users are developers and this is a precise message that allows to locate and fix the problem.

kopper:
> I think this PR is actually a part of "SD validation UX" ticket
> @kopper which solution would you prefer?

Why not both?

kopper:
> Are you suggesting we remove the schema and do all the validation by hand? =) I think schema should be used as much as possible. I think `Invalid JSON: locations[2].account is a required field` is pretty readable. Our users are developers and this is a precise message that allows to locate and fix the problem.

It is readable enough, but for most common error mistakes we should have even more readable. In some more extreme cases "did you mean X?" if possible.

ash2k:
> It is readable enough, but for most common error mistakes we should have even more readable. In some more extreme cases "did you mean X?" if possible.

If we really want to go down that path, we need to come up with an approach that would do this **consistently across the board**. I really don't want us to have a single snowflake case where we intentionally have weak schema and do validation in code. I think schema should be as strict as possible. Schema is declarative and consumable by tools, unlike checks that are in code.

If we really want extra smarts in validation, it should be in a validating webhook that is executed before the schema check IIRC. That is one approach off the top of my head that maybe can work.

kopper:
Two separate issues:
1. semantic validations, i.e. stuff schema can't validate (or it is awkward to do)
2. more user friendly error messages

AFAIU for 1 we need webhook

2 is part of UX and *ideally* should not be done in the API, but I don't think there is a way to extend kubectl with custom error messages so in practice we have two options:
1. implement custom error messages in webhook to cover both kubectl and atlas
2. assume users of kubectl are advanced users and dont need sophisticated messages ;-)

I prefer 1

scottgreenup:
I also think we should be consistent and it should mostly be achieved by schema validation. We could even write our own schema validation that detects that paths that have errors and adds extra help messages. But for now, this is better than not having schema validation in place.

nilebox:
Merging as is, since it's an improvement over the current story.
There is a separate epic for "SD validation story", and as part of it we should improve our JSON schema, agree with @ash2k .

PR #126, opened 2019-02-06
DENG-2197 trebuchet apiserver skeleton

qunayu:
Copy and pasted (what I thought was) Apiserver boilerplate code. Very likely that I've missed/copied something wrongly. Please review! 🙏 

https://hello.atlassian.net/browse/DENG-2197


pkg/apis/trebuchet/install/install.go
```diff
@@ -0,0 +1,12 @@
+package install
+
+import (
+	"github.com/atlassian/voyager/pkg/apis/creator/v1"
```
ash2k: This should be trebuchet api group
```suggestion
	trebuchet_v1 "github.com/atlassian/voyager/pkg/apis/trebuchet/v1"
```

pkg/apis/trebuchet/v1/types.go
```diff
@@ -0,0 +1,31 @@
+package v1
+
+import (
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ServiceResourceVersion = "v1"
```
ash2k: `ReleaseResourceVersion`

pkg/apis/trebuchet/v1/types_test.go
```diff
@@ -0,0 +1,9 @@
+package v1
+
+import (
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var _ runtime.Object = &Release{}
+var _ meta_v1.ObjectMetaAccessor = &Release{}
```
ash2k: Add
```go
var _ runtime.Object = &ReleaseList{}
var _ meta_v1.ListMetaAccessor = &ReleaseList{}
```

pkg/trebuchet/server/apiserver/apiserver.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	s := &TrebuchetServer{
+		GenericAPIServer: genericServer,
+	}
+
+	apiGroupInfo := genericapiserver.NewDefaultAPIGroupInfo(apis_trebuchet.GroupName, Scheme, metav1.ParameterCodec, Codecs)
+
```
nilebox: leave a TODO comment here? This is where you will inject a custom Store or HTTP handler to proxy request to underlying service instead of storing in etcd like standard API server.

pkg/trebuchet/server/options/trebuchet_options.go
```diff
@@ -0,0 +1,20 @@
+package options
+
+import (
+	"github.com/atlassian/ctrl"
+	"github.com/atlassian/ctrl/options"
+)
+
+type TrebuchetOptions struct {
+	LoggingOptions options.LoggerOptions
```
nilebox: TODO: these options are not used anywhere. You can leave a TODO comment for now if you prefer to add this code later.

pkg/trebuchet/server/rest/reststorage.go
```diff
@@ -0,0 +1,72 @@
...
+
+	trebuchet_v1 "github.com/atlassian/voyager/pkg/apis/trebuchet/v1"
+	"go.uber.org/zap"
+	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apiserver/pkg/registry/rest"
+)
+
+type REST struct {
```
nilebox: This struct is not used anywhere. See how it's injected in creator: https://github.com/atlassian/voyager/blob/master/pkg/creator/server/apiserver/apiserver.go#L92-L95

pkg/trebuchet/testdata/service_create.json
```diff
@@ -0,0 +1,11 @@
+{
+	"apiVersion": "creator.voyager.atl-paas.net/v1",
```
nilebox: This YAML is specific to Creator, you should have some example of Release instead (if you plan to test with it)

pkg/trebuchet/types.go
```diff
@@ -0,0 +1,14 @@
+package trebuchet
+
+import (
+	"net/http"
+
+	"github.com/atlassian/voyager/pkg/util/pkiutil"
+	"go.uber.org/zap"
+)
+
+type ExtraConfig struct {
```
nilebox: This should be constructed via `ApplyTo` method, currently isn't AFAICT.
See https://github.com/atlassian/voyager/blob/master/pkg/creator/server/options/creator_options.go#L35-L59

pkg/trebuchet/server/options/trebuchet_options.go
qunayu: Would it be a bad idea to leave this here, since it's likely that we'll need logging features in the future?

nilebox: @qunayu not sure what you mean. The logging options are definitely useful, they are just not used at the moment, so we need to make sure we don't forget to use them :)

qunayu: Ahh, I see. Thanks!

pkg/apis/trebuchet/v1/register.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // SchemeGroupVersion is group version used to register these objects.
-var SchemeGroupVersion = schema.GroupVersion{Group: trebuchet.GroupName, Version: ServiceResourceVersion}
+var SchemeGroupVersion = schema.GroupVersion{Group: trebuchet.GroupName, Version: "v1"}
```
ash2k: Why not use the constant `ReleaseResourceVersion`?

pkg/trebuchet/server/apiserver/apiserver.go
```diff
@@ -77,6 +83,15 @@ func (c completedConfig) New() (*TrebuchetServer, error) {
...
 	apiGroupInfo := genericapiserver.NewDefaultAPIGroupInfo(apis_trebuchet.GroupName, Scheme, metav1.ParameterCodec, Codecs)
 
+	// TODO: inject a custom Store or HTTP handler to proxy request to underlying service
+
+	v1Storage := map[string]rest.Storage{}
+	v1Storage["services"] = &trebuchetrest.REST{
+		Logger: c.ExtraConfig.Logger,
+	}
+
+	apiGroupInfo.VersionedResourcesStorageMap["v1"] = v1Storage
```
ash2k: Use the version constant

pkg/trebuchet/server/apiserver/apiserver.go
```diff
@@ -0,0 +1,101 @@
...
+	s := &TrebuchetServer{
+		GenericAPIServer: genericServer,
+	}
+
+	apiGroupInfo := genericapiserver.NewDefaultAPIGroupInfo(apis_trebuchet.GroupName, Scheme, metav1.ParameterCodec, Codecs)
+
+	// TODO: inject a custom Store or HTTP handler to proxy request to underlying service
+
+	v1Storage := map[string]rest.Storage{}
+	v1Storage["services"] = &trebuchetrest.REST{
```
nilebox: should be `"releases"` probably? (It's the name of resource type ("kind") that you want to handle). Trebuchet manages "releases" AFAIK.

pkg/trebuchet/server/options/trebuchet_options.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/atlassian/ctrl"
+	"github.com/atlassian/ctrl/options"
+	"github.com/atlassian/voyager/pkg/trebuchet"
+	"github.com/atlassian/voyager/pkg/util"
+	"github.com/atlassian/voyager/pkg/util/pkiutil"
+)
+
+// TODO: Add logging options for trebuchet
```
nilebox: nit: LoggingOptions are already there?..

PR #127, opened 2019-02-06
VYGR-308 prevent user error from bubbling up to ctrl (part 1)

ychen-atlassian:
This is the first part of a PR that tries to add some new things into orchestration to distinguish between external and non external errors.

### DIFF
--- a/pkg/orchestration/controller.go
+++ b/pkg/orchestration/controller.go
@@ -47,8 +47,8 @@ func ByConfigMapNameIndexKey(namespace string, configMapName string) string {
 }
 
 type Entangler interface {
-	Entangle(*orch_v1.State, *wiring.EntangleContext) (*smith_v1.Bundle, bool /*retriable*/, error)
-	Status(*orch_v1.StateResource, *wiring.StatusContext) (orch_v1.ResourceStatusData, bool /*retriable*/, error)
+	Entangle(*orch_v1.State, *wiring.EntangleContext) wiring.EntangleResult
+	Status(*orch_v1.StateResource, *wiring.StatusContext) wiring.EntangleStatusResult
 }
 
 type Controller struct {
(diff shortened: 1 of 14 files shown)


pkg/orchestration/controller.go
```diff
@@ -85,65 +85,105 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (retriable bool, err erro
...
-	conflict, retriable, err = c.handleProcessResult(ctx.Logger, state, bundle, retriable, err)
+	conflict, processRetriable, processErr := c.handleProcessResult(ctx.Logger, state, bundle, retriable, err)
 	if conflict {
 		return false, nil
 	}
+	if processErr != nil {
+		if err != nil && !external {
+			ctx.Logger.Info("Failed to set State status", zap.Error(processErr))
+			return processRetriable || retriable, err
+		} else if external {
```
ash2k: `err != nil && external` ? i.e. `external` has a valid value only if `err` is set.

ash2k:
```go
if err != nil {
    if external {
    } else {
    }
}
```
reads better?

pkg/orchestration/controller.go
```diff
@@ -85,65 +85,105 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (retriable bool, err erro
...
+			// We are going to return the error from handleProcessResult instead
+			ctx.Logger.Info("Invalid State Descriptor", zap.Error(err))
```
ash2k: `Invalid State object`?

pkg/orchestration/controller.go
```diff
@@ -85,65 +85,105 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (retriable bool, err erro
...
+		return retriable, err
+	}
+
+	return false, nil
 }
 
-// process processes the given State object performing autowiring for it.
-// It tries to return a Bundle even if there was an error.
-func (c *Controller) process(logger *zap.Logger, state *orch_v1.State) (conflictRet, retriableRet bool, b *smith_v1.Bundle, e error) {
+func (c *Controller) process(logger *zap.Logger, state *orch_v1.State) (b *smith_v1.Bundle, external bool, conflict bool, retriable bool, err error) {
```
ash2k: Please change it to not use named return values, use comments instead

p.s. this function is much cleaner now!

pkg/orchestration/controller.go
```diff
@@ -85,65 +85,105 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (retriable bool, err erro
...
+		return nil, errors.WithStack(err)
 	}
 	if !exists {
-		return false, false, nil, errors.Errorf("missing Namespace %q in informer", state.Namespace)
+		return nil, errors.Errorf("missing Namespace %q in informer", state.Namespace)
 	}
 	namespace := namespaceObj.(*core_v1.Namespace)
 
-	// Grab the ConfigMap
-	if state.Spec.ConfigMapName == "" {
```
ash2k: This should be in the schema

pkg/orchestration/controller.go
```diff
@@ -85,65 +85,105 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (retriable bool, err erro
...
+	}, nil
+}
+
+func (c *Controller) entangle(state *orch_v1.State, entangleContext *wiring.EntangleContext) (*smith_v1.Bundle, bool /* external */, bool /* retriable */, error) {
+	result := c.Entangler.Entangle(state, entangleContext)
+	switch r := result.(type) {
+	case *wiring.EntangleResultSuccess:
+		return r.Bundle, false, false, nil
+	case *wiring.EntangleResultFailure:
+		return nil, r.IsExternalError, r.IsRetriableError, errors.Wrapf(r.Error, "failed to wire up Bundle for State %q", state.Name)
```
ash2k: This message will be in the condition on the State object - no need to include it's name in the message

pkg/orchestration/controller.go
```diff
@@ -85,65 +85,105 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (retriable bool, err erro
...
+	default:
+		return nil, false, false, errors.Errorf("unknown entangler state %q", r.StatusType())
 	}
-	bundleSpec, retriable, err := c.Entangler.Entangle(state, entangleContext)
-	if err != nil {
-		return false, retriable, nil, errors.Wrapf(err, "failed to wire up Bundle for State %q", state.Name)
-	}
+}
 
+// process processes the given State object performing autowiring for it.
```
ash2k: Needs to be moved)

pkg/orchestration/controller.go
```diff
@@ -214,6 +254,10 @@ func copyCondition(bundle *smith_v1.Bundle, condType cond_v1.ConditionType, cond
 	}
 }
 
+// handleProcessResult takes the the results of processing and writes it into the
+// status of the State. Contrary to prior behavior it no longer returns
```
ash2k: Don't need to document the precious behavior =)

pkg/orchestration/controller.go
```diff
@@ -311,13 +354,16 @@ func (c *Controller) calculateResourceStatuses(stateResources []orch_v1.StateRes
 						Type:    cond_v1.ConditionError,
 						Status:  cond_v1.ConditionTrue,
 						Reason:  ReasonStatusRetrievalFailed,
-						Message: fmt.Sprintf("Failed to get status of resource: %v", err),
+						Message: fmt.Sprintf("Failed to get status of resource: %v", r.Error),
 					},
 				},
 			}
+		default:
```
ash2k: return error?

pkg/orchestration/wiring/aws/resources.go
```diff
@@ -35,9 +35,9 @@ const (
 // All osb-aws-provider resources are 'almost' the same, differing only in the service/plan names,
 // what they need passed in the ServiceEnvironment.
 var ResourceTypes = map[voyager.ResourceType]wiringplugin.WiringPlugin{
-	DynamoDB: wiringutil.StatusAdapter(Resource(DynamoDB, DynamoDBName, DynamoDBClass, DynamoDBPlan, dynamoDbServiceEnvironment, dynamoDbShapes).WireUp),
-	S3:       wiringutil.StatusAdapter(Resource(S3, S3Name, S3Class, S3Plan, s3ServiceEnvironment, s3Shapes).WireUp),
-	Cfn:      wiringutil.StatusAdapter(Resource(Cfn, CfnName, CfnClass, CfnPlan, CfnServiceEnvironment, cfnShapes).WireUp),
+	DynamoDB: wiringutil.TemporaryNewWiringMigrationAdapter(Resource(DynamoDB, DynamoDBName, DynamoDBClass, DynamoDBPlan, dynamoDbServiceEnvironment, dynamoDbShapes).WireUp),
+	S3:       wiringutil.TemporaryNewWiringMigrationAdapter(Resource(S3, S3Name, S3Class, S3Plan, s3ServiceEnvironment, s3Shapes).WireUp),
+	Cfn:      wiringutil.TemporaryNewWiringMigrationAdapter(Resource(Cfn, CfnName, CfnClass, CfnPlan, CfnServiceEnvironment, cfnShapes).WireUp),
```
ash2k: 😁

pkg/orchestration/wiring/entangler.go
```diff
@@ -135,14 +197,20 @@ func (en *Entangler) Entangle(state *orch_v1.State, context *EntangleContext) (*
 	for _, v := range sorted {
 		resource := g.Vertices[v].Data.(*orch_v1.StateResource)
 		dependants := getDependants(resource.Name, g.Vertices[v].IncomingEdges, state.Spec.Resources)
-		retriable, entErr := w.entangle(resource, &state.ObjectMeta, &stateContext, dependants)
+		external, retriable, entErr := w.entangle(resource, &state.ObjectMeta, &stateContext, dependants)
 		if entErr != nil {
-			return nil, retriable, errors.Wrapf(entErr, "failed to wire up resource %q of type %q", resource.Name, resource.Type)
+			return &EntangleResultFailure{
```
ash2k: As discussed earlier, it'd be great to not stop here but continue so that we detect more issues, not just the first one. Not in this PR, just saying. /cc @amckague

pkg/orchestration/controller.go
ash2k: Previous

pkg/orchestration/controller.go
ychen-atlassian: Yeah, this was in the old error message. I'll remove it.

pkg/orchestration/controller.go
ychen-atlassian: Yep - I deleted this because it was already in the schema.

In fact, some of the wiring functions have checks which are already described in the schema too - I'll do those as part of another PR

PR #128, opened 2019-02-06
VYGR-425: Separate service central types into read and write

amckague:
ServiceOwner is only valid as when read. Updating the owner is no longer
supported, writing the owner during creation is deprecated.

### DIFF
--- a/pkg/execution/svccatadmission/internaldns.go
+++ b/pkg/execution/svccatadmission/internaldns.go
@@ -80,14 +80,14 @@ func InternalDNSAdmitFunc(ctx context.Context, microsServerClient microsServerCl
 
 	if serviceCentralData == nil {
 		reason := fmt.Sprintf(
-			"namespace service %q does not exist in Service Central - should be impossible",
+			"namespace service %q does not exist in Service Central",
 			serviceName)
 		return &admissionv1beta1.AdmissionResponse{
 			Allowed: false,
 			Result: &metav1.Status{
 				Message: reason,
-				Code:    http.StatusForbidden,
-				Reason:  metav1.StatusReasonForbidden,
+				Code:    http.StatusInternalServerError,
+				Reason:  metav1.StatusReasonInternalError,
 			},
 		}, nil
 	}
--- a/pkg/execution/svccatadmission/internaldns.go
+++ b/pkg/execution/svccatadmission/internaldns.go
@@ -78,6 +78,20 @@ func InternalDNSAdmitFunc(ctx context.Context, microsServerClient microsServerCl
 		return nil, errors.Errorf("error fetching service central data for serviceName %q", serviceName)
 	}
 
+	if serviceCentralData == nil {
+		reason := fmt.Sprintf(
+			"namespace service %q does not exist in Service Central - should be impossible",
+			serviceName)
+		return &admissionv1beta1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: reason,
+				Code:    http.StatusForbidden,
+				Reason:  metav1.StatusReasonForbidden,
+			},
+		}, nil
+	}
+
 	// types to store parallel domain owner fetches
 	type domainOwnership struct {
 		requestedDomain         string
--- a/pkg/servicecentral/types.go
+++ b/pkg/servicecentral/types.go
@@ -15,7 +15,6 @@ type ServiceOwner struct {
 
 type ServiceDataWrite struct {
 	ServiceUUID        *string     `json:"service_uuid,omitempty"`
-	CreationTimestamp  *string     `json:"creation_timestamp,omitempty"`
 	ServiceName        ServiceName `json:"service_name,omitempty"`
 	ServiceTier        int         `json:"service_tier,omitempty"`
 	Tags               []string    `json:"tags,omitempty"`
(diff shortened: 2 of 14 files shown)


pkg/execution/svccatadmission/internaldns.go
```diff
@@ -78,6 +78,20 @@ func InternalDNSAdmitFunc(ctx context.Context, microsServerClient microsServerCl
 		return nil, errors.Errorf("error fetching service central data for serviceName %q", serviceName)
 	}
 
+	if serviceCentralData == nil {
```
amckague: @marciogh please focus on this change

pkg/servicecentral/types.go
```diff
@@ -35,6 +34,13 @@ type ServiceData struct {
 	Compliance *ServiceComplianceConf `json:"compliance,omitempty"`
 }
 
+type ServiceDataRead struct {
```
Burr: Compliance is also read-only, should move that out of Write.

pkg/execution/svccatadmission/internaldns.go
nilebox: Did we run into this issue recently? This could only happen when someone delete a service from Service Central without deleting SDs from clusters first.

nilebox: if `getServiceData` uses our ServiceCentral client, it returns an error for 404, not nil. So you could check for `if servicecentral.IsNotFound(err)` or similar.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -78,6 +78,20 @@ func InternalDNSAdmitFunc(ctx context.Context, microsServerClient microsServerCl
...
+		reason := fmt.Sprintf(
+			"namespace service %q does not exist in Service Central - should be impossible",
```
nilebox: "should be impossible" is misleading. As I commented above, someone deleted a service from Service Central, which is totally possible, but breaks our system.

pkg/execution/svccatadmission/internaldns.go
```diff
@@ -78,6 +78,20 @@ func InternalDNSAdmitFunc(ctx context.Context, microsServerClient microsServerCl
...
+			serviceName)
+		return &admissionv1beta1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: reason,
+				Code:    http.StatusForbidden,
```
nilebox: Worth double checking: we had an issue recently when webhook would reject request which deleted finalizer from the object (i.e. GC was proceeding with deletion). If we ran into the state of missing Service Central record, we should make sure that we don't block the deletion?

nilebox: Also, we could probably return 422 code instead here?

pkg/servicecentral/types.go
nilebox: OpsGenie Team is readonly too /cc @fcobb

pkg/servicecentral/types.go
```diff
@@ -13,18 +13,17 @@ type ServiceOwner struct {
...
-	ServiceTier        int          `json:"service_tier,omitempty"`
-	Tags               []string     `json:"tags,omitempty"`
-	Platform           string       `json:"platform,omitempty"`
-	Misc               []miscData   `json:"misc,omitempty"`
-	PagerDutyServiceID string       `json:"pagerduty_service_id,omitempty"`
-	LoggingID          string       `json:"logging_id,omitempty"`
-	SSAMContainerName  string       `json:"ssam_container_name,omitempty"`
+type ServiceDataWrite struct {
+	ServiceUUID        *string     `json:"service_uuid,omitempty"`
+	CreationTimestamp  *string     `json:"creation_timestamp,omitempty"`
```
nilebox: This looks readonly too

pkg/servicecentral/types.go
```diff
@@ -83,6 +89,7 @@ type V2Service struct {
 	Name string `json:"name" sql:"service_name,unique"`
 
 	// Owner of the service, must be an a valid staff ID of a person
+	// Owner is READ ONLY it cannot be set when talking to service central
```
nilebox: This comment seems redundant given that this is a response type, not request.

nilebox: Ignore me, GitHub diff looked like it was part of `serviceTypeResponse`

pkg/execution/svccatadmission/internaldns.go
amckague: The code doesn't actually handle this case correctly. If we do hit this we npe and block all changes to ingress for everyone.

pkg/execution/svccatadmission/internaldns.go
amckague: GC is covered by shared registration code for the webhooks. 422 sounds fine.

pkg/execution/svccatadmission/internaldns.go
amckague: The `getServiceData` function use list with a query. That would can return 0 results.

marciogh: Starting to look at this now

marciogh: https://sdog.jira-dev.com/browse/MICROS-6941

marciogh: @amckague @nilebox sorry, I don't see what's wrong here, need advice:

> if `getServiceData` uses our ServiceCentral client, it returns an error for 404, not nil. So you could check for `if servicecentral.IsNotFound(err)` or similar.

If I can't find the service on SC, I can't validate the domain, therefore, can't allow it to be created. So there's no real difference if I check for error or if serviceData is nil.

> The `getServiceData` function use list with a query. That would can return 0 results. (and a return result of nil, nil)

If I get nil, nil means the service couldn't be found. Again, I have to fail because don't have required information to allow the ingress.

>The code doesn't actually handle this case correctly. If we do hit this we npe and block all changes to ingress for everyone.

Where exactly the npe would happen?

PR #130, opened 2019-02-07
MICROS-6558-datadog-wiring

ssharifnabavi:
This is a wire up function to create datadog cpu and memory alarms


pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+func WireUp(stateResource *orch_v1.StateResource, context *wiringplugin.WiringContext) (*wiringplugin.WiringResult, bool, error) {
+	err := validateRequest(stateResource, context)
+	if err != nil {
+		return nil, false, err
+	}
+	kubeComputeDependency, err := context.TheOnlyDependency()
+	if err != nil {
+		return nil, false, err
+	}
+	setOfScalingShape, found, err := knownshapes.FindSetOfPScalingShape(kubeComputeDependency.Contract.Shapes)
```
ychen-atlassian: PScaling?

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	resourceName, err := oap.ResourceName(resource.Spec)
+	if err != nil {
+		return smith_v1.Resource{}, err
+	}
+	if resourceName == "" {
+		resourceName = string(resource.Name)
+	}
+
+	resourceName = resourceName + "--" + string(alarmType)
```
ychen-atlassian: You don't need to mutate resourceName, just use it directly in smithResourceName

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+		Region:      context.StateContext.Location.Region,
+	}
+
+	serviceInstanceSpecBytes, err := json.Marshal(&serviceInstanceSpec)
+	if err != nil {
+		return smith_v1.Resource{}, err
+	}
+	smithResourceName := voyager.ResourceName(resourceName)
+	alarmInstanceResource := smith_v1.Resource{
+		Name: wiringutil.ServiceInstanceResourceName(smithResourceName),
```
ychen-atlassian: This resource name should probably not be constructed from `oap.ResourceName()` because it could potentially clash. I think leaving resource.Name as the prefix is fine

ychen-atlassian: Also given none of our other wiring functions produce multiple service instances maybe you can create another function to handle this specific case in wiringutil? `ServiceInstanceWithPostfixResourceName` maybe?

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+	serviceInstanceSpec := ServiceInstanceSpec{
+		ServiceName: context.StateContext.ServiceName,
+		Attributes:  alarmsAtt,
+		Environment: context.StateContext.Location.EnvType,
+		Region:      context.StateContext.Location.Region,
+	}
+
+	serviceInstanceSpecBytes, err := json.Marshal(&serviceInstanceSpec)
+	if err != nil {
+		return smith_v1.Resource{}, err
```
ychen-atlassian: errors.WithStack

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Attributes  Alarm               `json:"attributes"`
+	Environment voyager.EnvType     `json:"environment"`
+	Region      voyager.Region      `json:"region"`
+}
+type AlarmOption struct {
+	Timeout           int             `json:"timeout_h,omitempty"`
+	locked            bool            `json:"locked,omitempty"`
+	NotifyNOData      bool            `json:"notify_no_data,omitempty"`
+	NodDataTimeFrame  string          `json:"no_data_timeframe,omitempty"`
```
ychen-atlassian: `NodData` typo

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+type ServiceInstanceSpec struct {
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Attributes  Alarm               `json:"attributes"`
+	Environment voyager.EnvType     `json:"environment"`
+	Region      voyager.Region      `json:"region"`
+}
+type AlarmOption struct {
+	Timeout           int             `json:"timeout_h,omitempty"`
+	locked            bool            `json:"locked,omitempty"`
+	NotifyNOData      bool            `json:"notify_no_data,omitempty"`
```
ychen-atlassian: NotifyNOData caps for some reason on NO

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+
+type ServiceInstanceSpec struct {
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Attributes  Alarm               `json:"attributes"`
+	Environment voyager.EnvType     `json:"environment"`
+	Region      voyager.Region      `json:"region"`
+}
+type AlarmOption struct {
+	Timeout           int             `json:"timeout_h,omitempty"`
+	locked            bool            `json:"locked,omitempty"`
```
ychen-atlassian: Did you want this as private?

pkg/orchestration/wiring/datadog/plugin.go
brjacobsatlassian: The 'P' is silent :trollface:

pkg/orchestration/wiring/datadog/plugin.go
brjacobsatlassian: I think there's a helper for this type of suffixing anyhow

pkg/orchestration/wiring/datadog/types.go
brjacobsatlassian: It would be weird to have a private serialized field, I think it's a typo

pkg/orchestration/wiring/wiringutil/knownshapes/scaling.go
```diff
@@ -0,0 +1,61 @@
...
+type SetOfScaling struct {
+	wiringplugin.ShapeMeta `json:",inline"`
+	Data                   SetOfScalingData `json:"data"`
+}
+
+// +k8s:deepcopy-gen=true
+type Scaling struct {
+	MinReplicas int32 `json:"minReplicas,omitempty"`
+	MaxReplicas int32 `json:"maxReplicas,omitempty"`
+	CPULimit    int32 `json:"cpuLimit,omitempty"`
```
kopper: remove these, they are unused

pkg/orchestration/wiring/wiringutil/knownshapes/scaling.go
```diff
@@ -0,0 +1,61 @@
+package knownshapes
+
+import (
+	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
+	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin"
+	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringutil/libshapes"
+)
+
+const (
+	// allows to select set of Kubernetes objects by list of scaling
```
kopper: update the comment

pkg/orchestration/wiring/wiringutil/knownshapes/scaling.go
```diff
@@ -0,0 +1,61 @@
...
+	SetOfScalingShape wiringplugin.ShapeName = "voyager.atl-paas.net/SetOfScalingShape"
```
kopper: looking for a better name...
ReplicatedShape?
(after you remove limits)

pkg/orchestration/wiring/testdata/k8scompute.datadog.bundle.yaml
```diff
@@ -0,0 +1,221 @@
...
+                  critical: 90
+                  warning: 80
+              query: avg(last_5m):( avg:kubernetes.cpu.usage.total{env:testenv,kube_namespace:default123,kube_deployment:state1,region:testregion}
+                by {container_id} / ( avg:kubernetes.cpu.limits{env:testenv,kube_namespace:default123,kube_deployment:state1,region:testregion}
+                by {container_id} * 1000000 ) ) * 100 > 90
+              type: metric alert
+            environment: testenv
+            region: testregion
+            serviceName: test-servicename
+  - name: alarms--memory--instance
```
kopper: do you plan to add alarm for too few replicas?

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return alarmInstanceResource, nil
+}
+
+func generateQuerySpec(context *wiringplugin.WiringContext, alarmType AlarmType, threshold AlarmThresholds) QueryParams {
+	return QueryParams{
+		Env:            string(context.StateContext.Location.EnvType),
+		Region:         string(context.StateContext.Location.Region),
+		KubeDeployment: string(context.StateMeta.Name),
+		KubeNamespace:  string(context.StateMeta.Namespace),
```
ash2k: These two above are strings already, no need to cast.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+			},
+		},
+	}
+	return alarmInstanceResource, nil
+}
+
+func generateQuerySpec(context *wiringplugin.WiringContext, alarmType AlarmType, threshold AlarmThresholds) QueryParams {
+	return QueryParams{
+		Env:            string(context.StateContext.Location.EnvType),
+		Region:         string(context.StateContext.Location.Region),
```
ash2k: Would be cleaner to change the fields to have the types from `Location.*`.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+				},
+			},
+		},
+	}
+	return alarmInstanceResource, nil
+}
+
+func generateQuerySpec(context *wiringplugin.WiringContext, alarmType AlarmType, threshold AlarmThresholds) QueryParams {
+	return QueryParams{
+		Env:            string(context.StateContext.Location.EnvType),
```
ash2k: `Env`->`EnvType` for consistency?

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type ServiceInstanceSpec struct {
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Attributes  Alarm               `json:"attributes"`
+	Environment voyager.EnvType     `json:"environment"`
+	Region      voyager.Region      `json:"region"`
+}
+type AlarmOption struct {
+	Timeout           int             `json:"timeout_h,omitempty"`
```
ash2k: 1. Would is be possible to change the field names to camelCase?
2. Don't use `int`, use `int32` or `int64`. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types
3. `TimeoutSeconds`. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units

pkg/orchestration/wiring/wiringutil/knownshapes/scaling.go
```diff
@@ -0,0 +1,61 @@
...
+			ShapeName: SetOfScalingShape,
+		},
+		Data: SetOfScalingData{
+			DeploymentResourceName: resourceName,
+			Scaling:                scaling,
+		},
+	}
+}
+
+func (s *SetOfScaling) Name() wiringplugin.ShapeName {
```
ash2k: This method can be remove - it is implemented by the embedded `wiringplugin.ShapeMeta`.

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+	Type    string      `json:"type"`
+	Query   string      `json:"query"`
+	Option  AlarmOption `json:"options"`
+	Message string      `json:"message"`
+}
+
+type ServiceInstanceSpec struct {
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Attributes  Alarm               `json:"attributes"`
+	Environment voyager.EnvType     `json:"environment"`
```
ash2k: Can we change it to `EnvType` and `envType` in json?

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+	serviceInstanceSpecBytes, err := json.Marshal(&serviceInstanceSpec)
+	if err != nil {
+		return smith_v1.Resource{}, err
+	}
+	smithResourceName := voyager.ResourceName(resourceName)
+	alarmInstanceResource := smith_v1.Resource{
+		Name: wiringutil.ServiceInstanceResourceName(smithResourceName),
+		References: []smith_v1.Reference{
+			{
+				Resource: deploymentResourceName,
```
ash2k: So the only purpose of `deploymentResourceName` is to ensure ordering of creation? We don't need any information from the kube compute resource?

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+	resourceName, err := oap.ResourceName(resource.Spec)
+	if err != nil {
+		return smith_v1.Resource{}, err
+	}
+	if resourceName == "" {
+		resourceName = string(resource.Name)
+	}
+
+	resourceName = resourceName + "--" + string(alarmType)
+	serviceInstanceSpec := ServiceInstanceSpec{
```
ash2k: Is this really enough to create an alarm? `Label` and resource name of the compute are not taken into account here. i.e. will it work if we have multiple compute resources in SD?

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+const (
+	clusterServiceClassExternalID                      = "875d4a87-e887-4838-a0b5-b64491dbf9cb"
+	clusterServicePlanExternalID                       = "d8048a2d-49de-4fda-b7ef-328de171cd32"
+	ResourceType                  voyager.ResourceType = "datadog-alarm"
+)
+
+type WiringPlugin struct {
+	svccatentangler.SvcCatEntangler
```
nilebox: Sorry for a late suggestion, but we have recently introduced a cleaner way of using common code for OSB autowiring, and plan to migrate all plugins to it and get rid of SvcCatEntangler.
See https://github.com/atlassian/voyager/pull/73 and https://github.com/atlassian/voyager/pull/74/files as examples

If you prefer to finish the PR still using SvcCatEntangler that's fine, I will rewrite it later.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+	sc_v1b1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
+	"github.com/pkg/errors"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	clusterServiceClassExternalID                      = "875d4a87-e887-4838-a0b5-b64491dbf9cb"
+	clusterServicePlanExternalID                       = "d8048a2d-49de-4fda-b7ef-328de171cd32"
+	ResourceType                  voyager.ResourceType = "datadog-alarm"
```
nilebox: `"DatadogAlarm"` per naming convention

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+type WiringPlugin struct {
+	svccatentangler.SvcCatEntangler
+}
+
+var (
+	serviceInstanceGVK = sc_v1b1.SchemeGroupVersion.WithKind("ServiceInstance")
```
nilebox: use constant, `WithKind(k8s.ServiceInstanceKind)`
or even better define a constant `k8s.ServiceInstanceGVK` (there are already examples, e.g. https://github.com/atlassian/voyager/blob/master/pkg/k8s/k8s.go#L68)

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, false, err
+	}
+	if !found {
+		return nil, false, errors.Errorf("failed to find shape %q in contract of %q", knownshapes.SetOfScalingShape, kubeComputeDependency.Name)
+	}
+
+	var wiredResources []smith_v1.Resource
+
+	deploymentResourceName := setOfScalingShape.Data.DeploymentResourceName
+	CPUServiceInstance, err := constructServiceInstance(stateResource, context, deploymentResourceName, CPU)
```
nilebox: `cpuServiceInstance`

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+
+type Alarm struct {
+	Name    string      `json:"name"`
+	Type    string      `json:"type"`
+	Query   string      `json:"query"`
+	Option  AlarmOption `json:"options"`
+	Message string      `json:"message"`
+}
+
+type ServiceInstanceSpec struct {
```
nilebox: This is a bad name for type (I see that it's the same in OAP, but still), because there is already ServiceInstanceSpec type in Service Catalog API, and this struct will be marshalled into Service Catalog's `ServiceInstanceSpec.Parameters.Raw`.

I propose to rename it to just `InstanceParameters` or `OSBInstanceParameters`.

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"github.com/atlassian/voyager"
+)
+
+type AlarmSpecType string
+type AlarmType string
+
+const (
+	Metric AlarmSpecType = "metric alert"
```
nilebox: is this the real enum value? having whitespace in enum value seems weird... or maybe it's not a enum?

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+)
+const (
+	CPU    AlarmType = "cpu"
+	Memory AlarmType = "memory"
+)
+
+type Alarm struct {
```
nilebox: `AlarmAttributes`? based on this struct being used as a value for `attributes` section below

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+	Name    string      `json:"name"`
+	Type    string      `json:"type"`
+	Query   string      `json:"query"`
+	Option  AlarmOption `json:"options"`
+	Message string      `json:"message"`
+}
+
+type ServiceInstanceSpec struct {
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Attributes  Alarm               `json:"attributes"`
```
nilebox: either rename `Alarm` type to `AlarmAttributes`, or rename `Attributes` to `Alarm` (and in JSON as well), to make them semantically closer

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+const (
+	CPU    AlarmType = "cpu"
+	Memory AlarmType = "memory"
+)
+
+type Alarm struct {
+	Name    string      `json:"name"`
+	Type    string      `json:"type"`
+	Query   string      `json:"query"`
+	Option  AlarmOption `json:"options"`
```
nilebox: `Options`? (based on JSON property name)

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+	Message string      `json:"message"`
+}
+
+type ServiceInstanceSpec struct {
+	ServiceName voyager.ServiceName `json:"serviceName"`
+	Attributes  Alarm               `json:"attributes"`
+	Environment voyager.EnvType     `json:"environment"`
+	Region      voyager.Region      `json:"region"`
+}
+type AlarmOption struct {
```
nilebox: `AlarmOptions`

pkg/orchestration/wiring/datadog/types.go
nilebox: As far as I understand, each of the fields there are optional. To make it more explicit, I would consider making all of them pointers, i.e.
```
type AlarmOption struct {
  Timeout *int
  Locked *bool
  ...
}
```
But that may lead to more boilerplate :(

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+	Thresholds        AlarmThresholds `json:"thresholds,omitempty"`
+}
+
+type AlarmThresholds struct {
+	Critical int32 `json:"critical"`
+	Warning  int32 `json:"warning,omitempty"`
+}
+
+type QueryParams struct {
+	KubeDeployment string
```
nilebox: json annotations missing

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -0,0 +1,57 @@
...
+	NotifyAudit       bool            `json:"notify_audit,omitempty"`
+	RequireFullWindow bool            `json:"require_full_window,omitempty"`
+	NewHostDelay      int             `json:"new_host_delay,omitempty"`
+	EscalationMessage string          `json:"escalation_message,omitempty"`
+	RenotifyInterval  int             `json:"renotify_interval,omitempty"`
+	Thresholds        AlarmThresholds `json:"thresholds,omitempty"`
+}
+
+type AlarmThresholds struct {
+	Critical int32 `json:"critical"`
```
nilebox: the one below has `omitempty`, while this one doesn't. Is that intentional?

pkg/orchestration/wiring/wiringutil/knownshapes/scaling.go
nilebox: I believe we have a constant for prefix `voyager.atl-paas.net` and use it in many places already

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -15,43 +15,42 @@ const (
...
 }
 
 type QueryParams struct {
-	KubeDeployment string
-	KubeNamespace  string
-	Env            string
-	Region         string
-	Threshold      *AlarmThresholds
-	AlarmType      AlarmType
+	KubeDeployment string           `json:"kubeDeployment,omitempty"`
```
nilebox: why some JSON fields are camel cased and others are separated with underscore? e.g. `json:"new_host_delay,omitempty"` above.

nilebox:
@ash2k can you give another round of review please (not all comments are addressed AFAICT)? Sharo is going to be away for the rest of the week, and it would be nice to merge this PR if there are no critical unaddressed review comments.

ash2k:
There are a few things that should be fixed. I'll do some cleanups and leave some questions.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -38,57 +38,63 @@ func WireUp(stateResource *orch_v1.StateResource, context *wiringplugin.WiringCo
...
+		},
+		Location: context.StateContext.Location,
 	}
+	message := query.generateMessage(&context.StateContext.ServiceProperties.Notifications)
 	serviceInstanceSpec := OSBInstanceParameters{
 		ServiceName: context.StateContext.ServiceName,
-		Attributes:  alarmsAtt,
 		Environment: context.StateContext.Location.EnvType,
 		Region:      context.StateContext.Location.Region,
+		Label:       context.StateContext.Location.Label,
```
ash2k: I have added the label here because it should be supported eventually.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -97,10 +103,10 @@ func constructServiceInstance(resource *orch_v1.StateResource, context *wiringpl
 	}
 
 	alarmInstanceResource := smith_v1.Resource{
-		Name: createAlarmNamesForSmithResource(resource.Name, deploymentResourceName, alarmType),
+		Name: wiringutil.ConsumerProducerResourceNameWithPostfix(resource.Name, kubeCompute, string(alarmType)),
```
ash2k: I'm using pre-existing helper functions here - this autowiring function should probably be changed to allow depending on multiple computes (why not?) for them to get the same alarms. And then it becomes a necessity to include the compute name into the resource name here (the "producer") to avoid clashes. Can be done separately.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -110,7 +116,7 @@ func constructServiceInstance(resource *orch_v1.StateResource, context *wiringpl
 					APIVersion: sc_v1b1.SchemeGroupVersion.String(),
 				},
 				ObjectMeta: meta_v1.ObjectMeta{
-					Name: wiringutil.ServiceInstanceWithPostfixMetaName(resource.Name, string(alarmType)),
+					Name: wiringutil.ConsumerProducerMetaNameWithPostfix(resource.Name, kubeCompute, string(alarmType)),
```
ash2k: Same as above

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -128,51 +134,16 @@ func constructServiceInstance(resource *orch_v1.StateResource, context *wiringpl
 	return alarmInstanceResource, nil
 }
 
-func generateQuerySpec(context *wiringplugin.WiringContext, alarmType AlarmType, threshold AlarmThresholds, deploymentResourceName smith_v1.ResourceName) QueryParams {
```
ash2k: I have inlined a bunch of things. I think it ready much better when you don't have to jump from func to func to see the whole parameters block.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -194,13 +165,10 @@ func (q *QueryParams) generateMessage(notificationProp *meta_orch.Notifications)
...
-	nameList := strings.Join([]string{string(deploymentResourceName), string(alarmType)}, "--")
-	return wiringutil.ServiceInstanceWithPostfixResourceName(resourceName, nameList)
-
-}
-
-func createAlarmNameforDatadog(serviceName string, resourceName string, alarmType string, env string, label string) string {
-	nameList := []string{serviceName, resourceName, alarmType, env, label}
+func createAlarmNameforDatadog(serviceName voyager.ServiceName, envType voyager.EnvType, resourceName voyager.ResourceName, alarmType AlarmType, label voyager.Label) string {
+	nameList := []string{string(serviceName), string(envType), string(resourceName), string(alarmType)}
+	if label != "" {
```
ash2k: Label is optional so always appending it is ugly.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -194,13 +165,10 @@ func (q *QueryParams) generateMessage(notificationProp *meta_orch.Notifications)
...
 
-func createAlarmNamesForSmithResource(resourceName voyager.ResourceName, deploymentResourceName smith_v1.ResourceName, alarmType AlarmType) smith_v1.ResourceName {
-	nameList := strings.Join([]string{string(deploymentResourceName), string(alarmType)}, "--")
-	return wiringutil.ServiceInstanceWithPostfixResourceName(resourceName, nameList)
-
-}
-
-func createAlarmNameforDatadog(serviceName string, resourceName string, alarmType string, env string, label string) string {
-	nameList := []string{serviceName, resourceName, alarmType, env, label}
+func createAlarmNameforDatadog(serviceName voyager.ServiceName, envType voyager.EnvType, resourceName voyager.ResourceName, alarmType AlarmType, label voyager.Label) string {
```
ash2k: Why everything was a string? I changed it to use types.

pkg/orchestration/wiring/datadog/types.go
ash2k: Still an open q

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -16,18 +16,19 @@ const (
...
+	Options AlarmOptions  `json:"options"`
+	Message string        `json:"message"`
 }
 
 type OSBInstanceParameters struct {
 	ServiceName voyager.ServiceName `json:"serviceName"`
-	Attributes  AlarmAttributes     `json:"attributes"`
 	Environment voyager.EnvType     `json:"environment"`
 	Region      voyager.Region      `json:"region"`
+	Label       voyager.Label       `json:"label,omitempty"`
```
ash2k: Added label

pkg/orchestration/wiring/datadog/types.go
```diff
@@ -16,18 +16,19 @@ const (
...
+	Attributes  AlarmAttributes     `json:"attributes"`
```
ash2k: Moved to be the last one after service name and location information.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -194,13 +165,10 @@ func (q *QueryParams) generateMessage(notificationProp *meta_orch.Notifications)
...
+	nameList := []string{string(serviceName), string(envType), string(resourceName), string(alarmType)}
```
ash2k: Reordered - environment goes after service name because its the next significant thing.

pkg/orchestration/wiring/testdata/k8scompute.datadog.bundle.yaml
```diff
@@ -180,7 +180,7 @@ spec:
                 [[/is_warning_recovery]] [[#is_alert]] @{123123123123123 https://events.pagerduty.com/adapter/cloudwatch_sns/v1/12312312312312312312312312312312}
                 [[/is_alert]][[#is_alert_recovery]] @{123123123123123 https://events.pagerduty.com/adapter/cloudwatch_sns/v1/12312312312312312312312312312312}
                 [[/is_alert_recovery]]
-              name: test-servicename-alarms-cpu-testenv-
```
ash2k: No dangling `-` anymore.

pkg/orchestration/wiring/wiringutil/knownshapes/datadog.go
```diff
@@ -1,47 +0,0 @@
...
-
-import (
-	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
-	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin"
-	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringutil/libshapes"
-)
-
-const (
-	// allows to select set of Kubernetes objects by list of datadog
-	SetOfDatadogShape wiringplugin.ShapeName = "voyager.atl-paas.net/SetOfDatadogShape"
```
ash2k: What is a "set of datadog"?

pkg/orchestration/wiring/wiringutil/knownshapes/datadog.go
```diff
@@ -1,47 +0,0 @@
-package knownshapes
-
-import (
-	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
-	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin"
-	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringutil/libshapes"
-)
-
-const (
-	// allows to select set of Kubernetes objects by list of datadog
```
ash2k: ????

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,174 @@
...
+	if stateResource.Type != ResourceType {
+		return errors.Errorf("invalid resource type: %q", stateResource.Type)
+	}
+	if stateResource.Spec != nil {
+		return errors.New("default alarm does not accept any user parameters")
+	}
+	return nil
+}
+
+func (q *QueryParams) generateQuery() string {
```
ash2k: This doesn't take label into consideration. I think it should.

pkg/orchestration/wiring/datadog/plugin.go
```diff
@@ -0,0 +1,174 @@
...
+		return cpuUsageString + cpuLimitString
+	}
+
+	memoryUsageString := fmt.Sprintf("avg(last_5m):( avg:kubernetes.memory.usage.total{env:%s,kube_namespace:%s,kube_deployment:%s,region:%s} by {container_id} ", q.Location.EnvType, q.KubeNamespace, q.KubeDeployment, q.Location.Region)
+	memoryLimitString := fmt.Sprintf("/ ( avg:kubernetes.memory.limits{env:%s,kube_namespace:%s,kube_deployment:%s,region:%s} by {container_id} * 1000000 ) ) * 100 > %d", q.Location.EnvType, q.KubeNamespace, q.KubeDeployment, q.Location.Region, q.Threshold.Critical)
+	return memoryUsageString + memoryLimitString
+
+}
+
+func (q *QueryParams) generateMessage(notificationProp *meta_orch.Notifications) string {
```
ash2k: Label is not handled

nilebox:
Apparently `SvcCatEntangler` was useless (everything was implemented by Sharo on her own 🙂), also merged latest master.

pkg/orchestration/wiring/datadog/plugin.go
Burr: Won't the label be taken care of by virtue of being in the `Namespace`? Or is that strictly the Service Name, even if it's in a label?

ash2k: You are right, it's in the namespace. I wonder if instead of the namespace we should filter by service name + label.

pkg/orchestration/wiring/datadog/plugin.go
ash2k: Namespace or label should be included in the message, otherwise it is not clear which deployment the alert is caused by.

PR #132, opened 2019-02-08
DENG-2198 Added new cluster roles for release & releaseGroups

alexwei:
Added new cluster roles in Synchronization controller to include new API group for release and release group.

### DIFF
--- a/pkg/synchronization/clusterroles.go
+++ b/pkg/synchronization/clusterroles.go
@@ -16,15 +16,21 @@ import (
 )
 
 const (
-	crServiceModifyAccessFormat      = "paas:creator:service:%s:modify"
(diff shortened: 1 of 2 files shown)


ash2k:
Do we really need a new cluster role and cluster role binding? Can we instead use the existing ones? Just trying to minimize number of extra objects...

nilebox:
I think you just need to extend the `paas:service-namespace` cluster role in Goliath, that is referenced in https://github.com/atlassian/voyager/blob/master/pkg/synchronization/roles.go#L27

You shouldn't need a change in Voyager at all.

alexwei:
Will make changes on Goliath side.

PR #133, opened 2019-02-08
MICROS-6921 Fix pod disruption budget for low minReplicas

halcyonCorsair:
If there are < 3 replicas, the kubecompute pod disruption budget must be raised accordingly to prevent node drain issues.

### DIFF
--- a/pkg/orchestration/wiring/k8scompute/plugin.go
+++ b/pkg/orchestration/wiring/k8scompute/plugin.go
@@ -542,11 +542,11 @@ func buildPodDisruptionBudgetSpec(labelMap map[string]string, minReplicas int32)
 	var pdbPercentage string
 	switch minReplicas {
 	case 0, 1:
-		pdbPercentage = "100%"
+		pdbPercentage = "0%"
 	case 2:
 		pdbPercentage = "50%"
 	default:
-		pdbPercentage = "30%"
+		pdbPercentage = "66%"
 	}
 
 	return policy_v1.PodDisruptionBudgetSpec{
--- a/pkg/orchestration/wiring/k8scompute/plugin.go
+++ b/pkg/orchestration/wiring/k8scompute/plugin.go
@@ -46,8 +46,6 @@ const (
 	hpaPostfix              = "hpa"
 	bindingOutputRoleARNKey = "IAMRoleARN"
 
-	defaultPodDisruptionBudget = "30%"
-
 	// Default environment variable names
 	awsRegionKey   = "MICROS_AWS_REGION"
 	envTypeKey     = "MICROS_ENVTYPE"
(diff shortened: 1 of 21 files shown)


pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -539,11 +537,22 @@ func buildAntiAffinity(labelMap map[string]string) *core_v1.PodAntiAffinity {
 	}
 }
 
-func buildPodDisruptionBudgetSpec(labelMap map[string]string) policy_v1.PodDisruptionBudgetSpec {
+func buildPodDisruptionBudgetSpec(labelMap map[string]string, minReplicas int32) policy_v1.PodDisruptionBudgetSpec {
+	// Make sure we can drain nodes if the minReplicas is < 3
+	var pdbPercentage string
+	switch minReplicas {
+	case 0, 1:
+		pdbPercentage = "100%"
```
patrickshan: should be 0 here since it's "MinAvailable" ?

halcyonCorsair: Whoops, yes percentages were inverted.

halcyonCorsair: Fixed now.

amckague:
I've removed red members from this PR. This sort of change (twiddling the output number) should be ok to go through without our team directly shepherding it.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -539,11 +537,22 @@ func buildAntiAffinity(labelMap map[string]string) *core_v1.PodAntiAffinity {
...
+func buildPodDisruptionBudgetSpec(labelMap map[string]string, minReplicas int32) policy_v1.PodDisruptionBudgetSpec {
+	// Make sure we can drain nodes if the minReplicas is < 3
+	var pdbPercentage string
+	switch minReplicas {
+	case 0, 1:
+		pdbPercentage = "0%"
+	case 2:
+		pdbPercentage = "50%"
+	default:
+		pdbPercentage = "66%"
```
jokeyrhyme: Is it okay for more than half the Pods to be disrupted? Should we maybe have 33% for replicas >= 3 ? Or maybe even make the `case 2` the default, for 50% with replicas >= 2 ?
If these were EC2s, we'd be this conservative, but I don't know if we can be more disruptive with Pods, maybe?

halcyonCorsair: This is a poorly named variable.  I made this mistake myself.  We are configuring "minimum available" here, eg. we want at least 2/3rds of the nodes to be available when > 3 nodes.

I'll rename the variable to make this more clear.

PR #134, opened 2019-02-08
VYGR-308 refactor compute to return user errors

ychen-atlassian:
one of many PRs

This changes EC2Compute and KubeCompute to try to classify error types as external or user errors.


ash2k:
you have  a conflict

PR #136, opened 2019-02-08
VYGR-308 expose user errors from kubeingress, postgres and sqs

ychen-atlassian:
One of many PRs

### DIFF
--- a/pkg/orchestration/wiring/kubeingress/plugin.go
+++ b/pkg/orchestration/wiring/kubeingress/plugin.go
@@ -44,28 +44,38 @@ const (
 )
 
 // WireUp is the main autowiring function for KubeIngress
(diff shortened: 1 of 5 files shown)


pkg/orchestration/wiring/kubeingress/plugin.go
```diff
@@ -44,37 +44,45 @@ const (
 )
 
 // WireUp is the main autowiring function for KubeIngress
-func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) (*wiringplugin.WiringResultSuccess, bool /*retriable*/, error) {
+func WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) wiringplugin.WiringResult {
 
 	// Fail if the resource type is wrong
 	if resource.Type != ResourceType {
```
ash2k: I wonder if we should just drop this check from all autowiring functions... If we really want to have it, we can have a wrapper that does the check before invoking the actual function. If you decide to do it, do it separately, not in this PR.

PR #138, opened 2019-02-10
Postgres and RDS autowiring clenups

ash2k:
First commit just formats the pre-existing fixtures, ignore it. Somehow the keys in some objects were not sorted - were manually edited?


pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -75,43 +74,13 @@ func (p *WiringPlugin) WireUp(resource *orch_v1.StateResource, context *wiringpl
...
-		References: references,
-		Spec: smith_v1.ResourceSpec{
-			Object: serviceInstance,
-		},
-	}
-
-	shapes, err := shapes(resource, &serviceInstanceResource, context)
+	// if this postgres depends on Dedicated RDS, it means it should be created there instead on the Default RDS
+	// there should be only one RDS dependency
+	sharedDbDep, found, err := context.FindTheOnlyDependency()
```
ash2k: This functions now has a single place where we are looking for the only dependency.

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -123,65 +92,57 @@ func shapes(resource *orch_v1.StateResource, smithResource *smith_v1.Resource, c
...
 	}
-
-	// If the postgres has an RDS dependency with read replica enabled, it
-	// produces some additional environment variables for the read replica
-	var foundSharedDb []*knownshapes.SharedDb
-	for _, dep := range context.Dependencies {
-		sharedDb, found, err := knownshapes.FindSharedDbShape(dep.Contract.Shapes)
+	if found {
+		var sharedDbShape *knownshapes.SharedDb
+		sharedDbShape, found, err = knownshapes.FindSharedDbShape(sharedDbDep.Contract.Shapes)
```
ash2k: This functions now has a single place where we are looking for the shape we need.

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -229,40 +190,15 @@ func instanceParameters(resource *orch_v1.StateResource, context *wiringplugin.W
 	// Filter instanceId
 	delete(finalSpec, "instanceId")
 
-	if finalSpec == nil {
```
ash2k: This is never true because it is always initialized above to a non-nil value (unless there is an error in which case we exit earlier).

pkg/orchestration/wiring/testdata/ec2compute.customrdsreplica.bundle.yaml
```diff
@@ -66,10 +66,7 @@ spec:
               MicrosVpcId: vpc-1
   - name: db--instance
     references:
-    - example: myownrds
-      name: rds--instance-metadata-name
-      path: metadata.name
-      resource: rds--instance
+    - resource: rds--instance
```
ash2k: This is only for ordering.

pkg/orchestration/wiring/testdata/ec2compute.customrdsreplica.bundle.yaml
```diff
@@ -87,7 +84,7 @@ spec:
               env: microstestenv
             resource_name: db
             shareddb:
-              resource: '!{rds--instance-metadata-name}'
```
ash2k: It used to pull in the name of the `ServiceInstance` but now uses the name of the Resource in the State. Both are `rds` strings, but semantically not the same thing. This is to match what RDS autowiring function does.

nilebox (dismissed):

ash2k:
rebased

nilebox (approved):

PR #139, opened 2019-02-10
rewrite to make validation possible

amckague:
hack and slash attempt to make the code more useful, let's discuss

Our current controllers are an intermixed mess of concerns, one of them is updating the "object model" the rest are various forms of "because kubernetes". Producing some sort of dry run is a nightmare in this world so I think some degree of refactoring is required.

As is I do not intend to merge

### DIFF
--- a/cmd/validator/validate.go
+++ b/cmd/validator/validate.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	ctrlApp "github.com/atlassian/ctrl/app"
+	"github.com/atlassian/voyager/pkg/apis/composition/v1"
+	"github.com/atlassian/voyager/pkg/composition"
+	"github.com/atlassian/voyager/pkg/options"
+	"github.com/atlassian/voyager/pkg/util/crash"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/yaml"
+)
+
+func main() {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	defer crash.LogPanicAsJSON()
+	ctrlApp.CancelOnInterrupt(ctx, cancelFunc)
+
+	err := runWithContext(ctx)
+
+	if err != nil && errors.Cause(err) != context.Canceled {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func runWithContext(context context.Context) error {
+	sd, err := LoadSD("cmd/validator/testdata/sd.yaml")
+	if err != nil {
+		return err
+	}
+
+	// Composition
+	clusterLocation, err := LoadLocation("cmd/validator/testdata/location.yaml")
+	sdTransformer := composition.NewServiceDescriptorTransformer(clusterLocation.ClusterLocation())
+	formationObjectResults, err := composition.ProcessSD(sd, sdTransformer)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(Print(json.MarshalIndent(composition.Locations(formationObjectResults), "", "  ")))
+
+	return nil
+}
+
+func Print(b []byte, err error) string {
+	if err != nil {
+		return err.Error()
+	}
+	return string(b)
+}
+
+func LoadLocation(filename string) (*options.Location, error) {
+	file, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	bytes, err := ioutil.ReadAll(file)
+	if err != nil {
+		return nil, err
+	}
+	location := &options.Location{}
+	err = yaml.UnmarshalStrict(bytes, location)
+	if err != nil {
+		return nil, err
+	}
+	return location, nil
+}
+
+func LoadSD(filename string) (*v1.ServiceDescriptor, error) {
+	file, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	bytes, err := ioutil.ReadAll(file)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/composition/controller2.go
```diff
@@ -0,0 +1,624 @@
+package composition
```
amckague: Here's the rewrite. Assume it's new code your reviewing from scratch. Refer to the old code if you're curious about something but otherwise focus on the structure of this code.

pkg/composition/controller2.go
```diff
@@ -0,0 +1,624 @@
...
+			if info.Namespace != c.namespace && c.namespace != "" {
+				formLogger.Sugar().Infof("Namespace %q doesn't match namespace %q observed by the controller. Skipping formationObjectDef %v.",
+					info.Namespace, c.namespace, info.Location.Label)
+				return true
+			}
+			return false
+		},
+	}
+
+	formationObjectDefList, err := ProcessSD(sd, filteredTransformer)
```
amckague: Start by figuring out how what the new outputs should be.

pkg/composition/controller2.go
```diff
@@ -0,0 +1,624 @@
...
+	}
+
+	var (
+		results        []*formationObjectResult
+		retryable      = false
+		deleteFinished = true
+		processFailure error
+	)
+
+	for _, result := range foResults {
```
amckague: Attempt to create namespaces + lds

pkg/composition/controller2.go
```diff
@@ -0,0 +1,624 @@
...
+			deleteFinished = false
+			processFailure = processErr
+			retryable = processRetryable
+			break
+		}
+		results = append(results, newResult)
+	}
+
+	// If deletion is complete, remove finalizer
+	if sd.ObjectMeta.DeletionTimestamp != nil && deleteFinished {
```
amckague: Move the finalizer logic out of status calculation since it doesn't really have anything to do with updating the status and has more to do with patching over the delete story.

pkg/composition/controller2.go
```diff
@@ -0,0 +1,624 @@
...
+		if err != nil {
+			return false, errors.WithStack(err)
+		}
+
+		for _, obj := range existingLds {
+			existingLD[ns] = append(existingLD[ns], obj.(*form_v1.LocationDescriptor))
+		}
+	}
+
+	st := statusUpdater{
```
amckague: Define a smaller interface for figuring out the status update. This is more approachable as an interface to use for dry run/validation. This would allow status updates from formation or autowiring to "bubble" back up in a validator without needing the kube api.

pkg/composition/controller2.go
```diff
@@ -0,0 +1,624 @@
...
+		formLogger := logger.With(logz.NamespaceName(result.namespace.Namespace))
+		formLogger.Debug("Processing LocationDescriptor for ServiceDescriptor")
+
+		var (
+			newResult        *formationObjectResult
+			processConflict  bool
+			processRetryable bool
+			processErr       error
+		)
+		if sd.ObjectMeta.DeletionTimestamp != nil {
```
amckague: The original code treats updates and deletes as the same thing complicating the return result and flow control for that process. Here it's still pretty complicated. 

For example: what happens when the user removes a location?

pkg/composition/controller2.go
```diff
@@ -0,0 +1,624 @@
...
+		}
+	}
+
+	st := statusUpdater{
+		clock:              c.clock,
+		existingNamespaces: existingNamespaces,
+		existingLD:         existingLD,
+		location:           c.location,
+
+		conditionUpdate: func(name string, condition cond_v1.Condition) {
```
amckague: Could be a better way of instrumenting these changes, suggestions welcome.

cmd/validator/validate.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func runWithContext(context context.Context) error {
+	sd, err := LoadSD("cmd/validator/testdata/sd.yaml")
+	if err != nil {
+		return err
+	}
+
+	// Composition
+	clusterLocation, err := LoadLocation("cmd/validator/testdata/location.yaml")
```
nilebox: Not sure if I understood your idea correctly, but Kubernetes now supports "dry run" mode which we can use instead of having a separate binary?
See https://github.com/kubernetes/kubernetes/issues/11488
it's a server-side feature that is enabled via passing a flag inside create/update requests, e.g. https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L498-L504

Need to confirm that this flag is respected by `apiextensions-apiserver` (CRD).

amckague: Intention is to separate the code that does the useful work from the kubernetes code around it so we can, say, run validation without round tripping through 4 objects or 4 separate dry run calls.

nilebox: Sorry but I don't see this validator going through 4 objects. At the moment it does more or less the same as what the replication validation webhook does, i.e. taking SD and trying to transform it into "formation objects"? Do you plan to extend this validator to do the whole chain of SD->LD->State->Bundle conversion? That would also require location-specific config, service metadata, deployment artifacts and other things?

amckague: Need to rewrite all of the controllers to make it possible. 
> Do you plan to extend this validator to do the whole chain of SD->LD->State->Bundle conversion 

yes

> location-specific config, service metadata, deployment artifacts and other things?

yes

pkg/composition/controller2.go
ychen-atlassian: Yeah we don't handle deletions correctly at all.

PR #149, opened 2019-02-11
M2GOLD-12 Support external cluster registry in voyager aggregator

nickhiggs:
Threw together support for an external cluster registry stored in S3 or similar. Also added in some local config for the aggregator so it can be run outside a cluster for local dev.


pkg/util/clusterregistry/external.go
```diff
@@ -0,0 +1,69 @@
...
+		namespace:    "kube-system",
+		rest:         rest,
+		cache:        cache.NewThreadSafeStore(cache.Indexers{}, cache.Indices{}),
+		resyncPeriod: resyncPeriod,
+	}, nil
+
+}
+
+func (e *ExternalClusterRegistry) GetClusters() ([]*cr_v1a1.Cluster, error) {
+	if time.Since(e.lastRun) > e.resyncPeriod {
```
ash2k: - `ExternalClusterRegistry` is supposed to be thread-safe, right? Currently it is not because `e.lastRun` is read and written without any synchronization means.
- Should likely avoid making multiple concurrent calls to the registry because why if you can do a single call and then serve data to multiple callers.
- Since you are maintaining a cache here, why not make it asynchronous and decouple reads from cache refreshes? Have a goroutine that is periodically refreshing the cache and always serve data from it? Can even use watch if it is supported to refresh data in the cache. How sensitive are we to stale information in that cache?

pkg/util/clusterregistry/external.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+		}
+
+		e.cache.Replace(clusters, "")
+		e.lastRun = time.Now()
+	}
+
+	clusters := e.cache.List()
+
+	typedClusters := make([]*cr_v1a1.Cluster, 0)
```
ash2k: preallocate

pkg/util/clusterregistry/informer.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = cctx.RegisterInformer(clusterGVK, clusterInformer.informer)
+	return clusterInformer, err
+}
+
+func (c *ClusterInformer) GetClusters() ([]*cr_v1a1.Cluster, error) {
+	clusters, err := c.informer.GetIndexer().ByIndex(ByClusterLabelIndexName, "paas")
```
ash2k: paas should not be hardcoded, move to config file

pkg/util/clusterregistry/informer.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	return []string{clusterLabel}, nil
+}
+
+type ClusterInformer struct {
+	informer cache.SharedIndexInformer
+}
+
+func NewClusterInformer(config *ctrl.Config, cctx *ctrl.Context) (*ClusterInformer, error) {
```
ash2k: Ideally this package should not depend on `ctrl`. All the plumbing logic should be in the controller constructor, like in all other controllers.

pkg/aggregator/server/options/options.go
```diff
@@ -63,3 +74,28 @@ func (o *AggregatorServerOptions) Config() (*apiserver.Config, error) {
...
+	}
+	if err := opts.Features.ApplyTo(&config.Config); err != nil {
+		return err
+	}
+	if err := opts.CoreAPI.ApplyTo(config); err != nil {
+		return err
+	}
+	if initializers, err := opts.ExtraAdmissionInitializers(config); err != nil {
+		return err
+	} else if err := opts.Admission.ApplyTo(&config.Config, config.SharedInformerFactory, config.ClientConfig, scheme, initializers...); err != nil {
```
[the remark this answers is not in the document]
nickhiggs: Allows reuse of the initializers which is only valid in this context

Makefile
```diff
@@ -314,6 +314,24 @@ run-smith: validate-cluster
...
+run-aggregator:
+	KUBE_PATCH_CONVERSION_DETECTOR=true \
+	KUBE_CACHE_MUTATION_DETECTOR=true \
+	bazel run $(BAZEL_OPTIONS) \
+		//cmd/aggregator:aggregator_race \
+		-- \
+		--config "$(CURDIR)/build/local/aggregator/config.yaml" \
+		--client-config-from=file \
+		--client-config-file-name="$$HOME/.kube/config" \
+		--kubeconfig="$$HOME/.kube/config" \
```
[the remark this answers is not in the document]
nickhiggs: From memory this was because different apiserver config components require different config flags when not running in-cluster

PR #150, opened 2019-02-12
MICROSHELP 5618: Increase reprep resilience

scottgreenup:
Increase the resilience of reporter reporter.

### DIFF
--- a/pkg/reporterreporter/BUILD.bazel
+++ b/pkg/reporterreporter/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
@@ -16,4 +16,11 @@ go_library(
         "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
         "//vendor/k8s.io/client-go/kubernetes:go_default_library",
     ],
+)
+
+go_test(
+    name = "go_default_test",
+    srcs = ["report_test.go"],
+    embed = [":go_default_library"],
+    deps = ["//vendor/github.com/stretchr/testify/assert:go_default_library"],
 )
--- a/pkg/reporterreporter/BUILD.bazel
+++ b/pkg/reporterreporter/BUILD.bazel
@@ -8,7 +8,9 @@ go_library(
     deps = [
         "//pkg/apis/reporter/v1:go_default_library",
         "//pkg/reporter/client:go_default_library",
+        "//pkg/util:go_default_library",
         "//vendor/bitbucket.org/atlassianlabs/restclient:go_default_library",
+        "//vendor/github.com/atlassian/ctrl:go_default_library",
         "//vendor/github.com/pkg/errors:go_default_library",
         "//vendor/go.uber.org/zap:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
--- a/pkg/reporterreporter/BUILD.bazel
+++ b/pkg/reporterreporter/BUILD.bazel
@@ -9,6 +9,7 @@ go_library(
         "//pkg/apis/reporter/v1:go_default_library",
         "//pkg/reporter/client:go_default_library",
         "//pkg/util:go_default_library",
+        "//pkg/util/sets:go_default_library",
         "//vendor/bitbucket.org/atlassianlabs/restclient:go_default_library",
         "//vendor/github.com/atlassian/ctrl:go_default_library",
         "//vendor/github.com/pkg/errors:go_default_library",
--- a/pkg/reporterreporter/BUILD.bazel
+++ b/pkg/reporterreporter/BUILD.bazel
@@ -13,7 +13,9 @@ go_library(
         "//vendor/github.com/atlassian/ctrl:go_default_library",
         "//vendor/github.com/pkg/errors:go_default_library",
         "//vendor/go.uber.org/zap:go_default_library",
+        "//vendor/k8s.io/api/core/v1:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
+        "//vendor/k8s.io/apimachinery/pkg/util/wait:go_default_library",
         "//vendor/k8s.io/client-go/kubernetes:go_default_library",
     ],
 )
--- a/pkg/reporterreporter/BUILD.bazel
+++ b/pkg/reporterreporter/BUILD.bazel
@@ -20,7 +20,9 @@ go_library(
 
 go_test(
     name = "go_default_test",
+    size = "small",
     srcs = ["report_test.go"],
     embed = [":go_default_library"],
+    race = "on",
     deps = ["//vendor/github.com/stretchr/testify/assert:go_default_library"],
 )
--- a/pkg/reporterreporter/report.go
+++ b/pkg/reporterreporter/report.go
@@ -12,6 +12,7 @@ import (
 	"github.com/atlassian/voyager/pkg/util"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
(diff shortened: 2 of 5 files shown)


scottgreenup:
Some of the namespaces aren't for Micros2 services, and those error for reporterreporter, this will survive those for now. @ash2k How do you actually check if a namespace is for a Micros2 service?

pkg/reporterreporter/BUILD.bazel
```diff
@@ -8,10 +8,19 @@ go_library(
...
         "//vendor/bitbucket.org/atlassianlabs/restclient:go_default_library",
+        "//vendor/github.com/atlassian/ctrl:go_default_library",
         "//vendor/github.com/pkg/errors:go_default_library",
         "//vendor/go.uber.org/zap:go_default_library",
         "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1:go_default_library",
         "//vendor/k8s.io/client-go/kubernetes:go_default_library",
     ],
 )
+
+go_test(
```
ash2k: missing size and race flags

pkg/reporterreporter/report.go
```diff
@@ -6,21 +6,29 @@ import (
...
 )
 
+const (
+	retryAttemptsPerReport = 5
+	retryDelay             = time.Second * 1
+)
+
 type Report struct {
-	RemoteURI        string
 	Cluster          string
```
ash2k: You can just make all the fields private because now it's impossible to construct a valid instance anyway - `mutator` is private.

pkg/reporterreporter/report.go
```diff
@@ -30,49 +38,98 @@ type RequestData struct {
...
+	return statusCode >= 500 && statusCode <= 599
+}
+
+func (r *Report) sendData(requestData RequestData) (retriable bool, err error) {
+	req, err := r.mutator.NewRequest(restclient.BodyFromJSON(requestData))
+	if err != nil {
+		return false, errors.Wrap(err, "could not craft a HTTP request")
+	}
+
+	resp, err := r.HTTPClient.Do(req)
```
ash2k: need to close the response body.

pkg/reporterreporter/report.go
```diff
@@ -30,49 +38,98 @@ type RequestData struct {
...
-				return errors.Wrap(err, "failed to connect to backend")
-			}
-			if resp.StatusCode != http.StatusOK {
-				return errors.New("failed to POST the data")
+			for i := 1; i <= retryAttemptsPerReport; i++ {
+				if retriable, err := r.sendData(requestData); err != nil {
+					if !retriable {
+						// If the request is not retriable, then we shouldn't try the other requests either
+						return err
+					} else {
```
ash2k: remove else

ash2k:
> How do you actually check if a namespace is for a Micros2 service?

It has a label with service name on it.

pkg/reporterreporter/report.go
scottgreenup: Shouldn't we be making Report an interface instead? Making them all private doesn't change how the ability to create one?

I could also make mutator public?

pkg/reporterreporter/report.go
```diff
@@ -30,49 +38,98 @@ type RequestData struct {
...
+						r.Logger.Warn("Failed to send request data to slurper",
+							zap.Int("attempt", i),
+							zap.Int("max_attempts", retryAttemptsPerReport),
```
ychen-atlassian: I'd actually be okay with not having these fields. I think it's nice to restrict these to fields we actually want to filter on because if we use splunk the most common way is to look at the msg field and use the json fields as filtering (or have error/msg as two separate columns).

Plus the more fields we have here, the more clog we get in splunk

pkg/reporterreporter/report.go
```diff
@@ -30,49 +39,137 @@ type RequestData struct {
...
+		mutator: restclient.NewRequestMutator(
+			restclient.BaseURL(slurperURI),
+			restclient.JoinPath(slurperPath),
+			restclient.Method(http.MethodPost)),
+		reporterClient: reporterClient,
+	}
+}
+
+func isRetriableError(statusCode int) bool {
+	_, ok := map[int]bool{
```
amckague: This is a set `map[string]interface{} { http.StatusRequestTimeout: interface{}... ` roughly, rather than storing bools.

amckague: or the sets package we use

pkg/reporterreporter/report.go
```diff
@@ -30,49 +38,98 @@ type RequestData struct {
...
-				return errors.Wrap(err, "invalid request")
-			}
-			resp, err := r.HTTPClient.Do(req)
-			if err != nil {
-				return errors.Wrap(err, "failed to connect to backend")
-			}
-			if resp.StatusCode != http.StatusOK {
-				return errors.New("failed to POST the data")
+			for i := 1; i <= retryAttemptsPerReport; i++ {
+				if retriable, err := r.sendData(requestData); err != nil {
```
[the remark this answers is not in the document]
scottgreenup: I've given it a crack :-P

pkg/reporterreporter/report.go
```diff
@@ -30,49 +39,137 @@ type RequestData struct {
...
+			}
 		}
-		if len(list.Items) == 0 {
-			r.Logger.Sugar().Infof("%q is empty", namespace.Name)
-			continue
+
+		return false, errors.Wrap(err, "error occured during request to slurper")
+
+	}
+	defer resp.Body.Close()
```
amckague: should be on line 82

scottgreenup: https://sourcegraph.com/github.com/golang/go/-/blob/src/net/http/client.go#L480-484

pkg/reporterreporter/report.go
```diff
@@ -30,49 +41,143 @@ type RequestData struct {
...
+
+	if ok {
+		return true
+	}
+
+	return statusCode >= 500 && statusCode <= 599
+}
+
+// sendData will send the requestData containing the report to slurper
+func (r *report) sendData(requestData RequestData) (retriable bool, err error) {
```
amckague: Why not pass context down to this level and include a reference in the rest client?

pkg/reporterreporter/report.go
scottgreenup: Done.

pkg/reporterreporter/report.go
scottgreenup: I've added it; I didn't add it because it wasn't really being used, but I guess it's good for the future.

pkg/reporterreporter/report.go
```diff
@@ -30,49 +43,137 @@ type RequestData struct {
 	Data      reporter_v1.NamespaceReport `json:"data"`
 }
 
-func (r *Report) Run(ctx context.Context) error {
-	namespaces, err := r.KubernetesClient.CoreV1().Namespaces().List(meta_v1.ListOptions{})
+func NewReport(slurperURI string, cluster string, reporterClient client.Interface, kubeClient kubernetes.Interface, logger *zap.Logger) ctrl.Server {
+	logger.Info("ReporterReporter configured",
```
ash2k: We don't do this anywhere else, why here?

pkg/reporterreporter/report.go
```diff
@@ -30,49 +43,137 @@ type RequestData struct {
...
+		mutator: restclient.NewRequestMutator(
+			restclient.BaseURL(slurperURI),
+			restclient.JoinPath(slurperPath),
+			restclient.Method(http.MethodPost)),
+		reporterClient: reporterClient,
+	}
+}
+
+func isRetriableError(statusCode int) bool {
+	return sets.NewInt(
```
ash2k: This should be
```go
switch statusCode { 
   case http.StatusRequestTimeout, http.StatusTooManyRequests, http.StatusServiceUnavailable, http.StatusGatewayTimeout:
   return true
   default:
     return false
}
```

pkg/reporterreporter/report.go
```diff
@@ -30,49 +43,137 @@ type RequestData struct {
...
+				return true, errors.Wrap(err, "transient error occurred during request to slurper")
+			}
 		}
-		if len(list.Items) == 0 {
-			r.Logger.Sugar().Infof("%q is empty", namespace.Name)
+
+		return false, errors.Wrap(err, "error occured during request to slurper")
+
+	}
+	defer resp.Body.Close() // nolint: errcheck
```
ash2k: there is a helper in util to silently close it

pkg/reporterreporter/report.go
```diff
@@ -30,49 +43,137 @@ type RequestData struct {
...
+	return sets.NewInt(
+		http.StatusRequestTimeout,
+		http.StatusTooManyRequests,
+		http.StatusServiceUnavailable,
+		http.StatusGatewayTimeout).Has(statusCode)
+}
+
+// sendData will send the requestData containing the report to slurper
+func (r *report) sendData(ctx context.Context, requestData RequestData) (retriable bool, err error) {
+	req, err := r.mutator.NewRequest(restclient.Context(ctx), restclient.BodyFromJSON(requestData))
```
ash2k: Is this sending data via a GET request?

pkg/reporterreporter/report.go
```diff
@@ -30,49 +43,137 @@ type RequestData struct {
...
+
+	}
+	defer resp.Body.Close() // nolint: errcheck
+
+	// We don't need to check if resp is nil; it won't be if err is nil
+	if resp.StatusCode == http.StatusOK {
+		return false, nil
+	}
+
+	return isRetriableError(resp.StatusCode), errors.Errorf("sending data to slurper failed")
```
ash2k: No need for formatting

pkg/reporterreporter/report.go
```diff
@@ -30,49 +43,137 @@ type RequestData struct {
...
+}
+
+func IsServiceNamespace(namespace core_v1.Namespace) bool {
+	for k := range namespace.Labels {
```
ash2k: labels is a map, no need to iterate

pkg/reporterreporter/report.go
```diff
@@ -30,49 +43,137 @@ type RequestData struct {
...
+		if k == "voyager.atl-paas.net/serviceName" {
```
ash2k: There is a constant for this

PR #151, opened 2019-02-12
MICROS-6954 Change PlatformDNS to allow only one alias per resource

mcoot:
Modifies the PlatformDNS resource so that it can only create one alias, rather than a list of aliases.

This is a breaking change for existing ServiceInstances. It also has an associated provider-side change (also in PR, that will have to be merged before this change is deployed).


marciogh:
@ash2k yep, this resource is not in use yet. The only related piece of work that touch resources in production is the PlatformDNS webhook, which also validates Ingresses.

mcoot:
> This resource is not used by anyone, so migration is not needed, right?

Yes. There are a small number of test serviceinstances in dev that will need to be cleaned up, but it is not officially available yet.

PR #153, opened 2019-02-13
Formation user errors

ychen-atlassian:
This fixes various issues I found (things being logged at Info level rather than Error), plus some simplification of some logic.

It also adds the same change I made to composition and orchestration into formation to allow user errors to not be pulled into ctrl.


pkg/formation/controller_test.go
```diff
@@ -552,16 +594,38 @@ func TestProcessLocationDescriptorWithMultipleTemplatingErrorsReturnsErrorList(t
...
-			require.Equal(t, 3, numErrors) // One error per resource
-			for _, e := range err.(*util.ErrorList).ErrorList {
-				assert.Contains(t, e.Error(), "variable not defined")
-			}
+			require.NoError(t, err)
+			// numErrors := len(err.(*util.ErrorList).ErrorList)
+			// require.Equal(t, 3, numErrors) // One error per resource
+			// for _, e := range err.(*util.ErrorList).ErrorList {
+			// 	assert.Contains(t, e.Error(), "variable not defined")
+			// }
```
ash2k: remove comments

PR #154, opened 2019-02-13
VYGR-459: filter out services with no name

amckague:


amckague:
@ychen-atlassian lost your comment in the rebase :(

PR #161, opened 2019-02-13
VYGR-461: Migrate OAP to the new OSB wiring helper

nilebox:


pkg/orchestration/wiring/aws/plugin.go
```diff
@@ -38,21 +44,72 @@ func Resource(resourceType voyager.ResourceType,
...
+	}
+	if instanceParameters != nil {
+		serviceInstance.Spec.Parameters = &runtime.RawExtension{
+			Raw: instanceParameters,
+		}
+	}
+
+	instanceResourceName := wiringutil.ServiceInstanceResourceName(resource.Name)
+
+	smithResource := smith_v1.Resource{
```
ash2k: I find it more readable to merge all the struct literal into a single big one. But it's not a big deal.

pkg/orchestration/wiring/aws/plugin.go
```diff
@@ -38,21 +44,72 @@ func Resource(resourceType voyager.ResourceType,
 	shapes svccatentangler.ShapesFunc,
 ) *WiringPlugin {
 	wiringPlugin := &WiringPlugin{
```
ash2k: Can just return here directly.

PR #162, opened 2019-02-13
EDGE-891: global edge provider integration

nilebox:
Just moved the branch from the fork to Voyager repo, see original PR https://github.com/atlassian/voyager/pull/74


pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+
+type ResourceParameters struct {
+	Attributes Attributes `json:"attributes"`
+}
+
+type Attributes struct {
+	UpstreamAddress []UpstreamAddress `json:"upstream_address"`
+	UpstreamPort    int32             `json:"upstream_port,omitempty"`
+	UpstreamSuffix  string            `json:"upstream_suffix,omitempty"`
+	UpstreamOnly    string            `json:"upstream_only,omitempty"`
```
mmatusiak-atlassian: is this a bool?

pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+	Domain          []string          `json:"domain,omitempty"`
```
mmatusiak-atlassian: singular name for plural thing?

pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+	Healthcheck     string            `json:"healthcheck,omitempty"`
+	Rewrite         string            `json:"rewrite,omitempty"`
```
mmatusiak-atlassian: bool again?

pkg/orchestration/wiring/edge/types.go
nilebox: this is an existing OSB contract (https://stash.atlassian.com/projects/MICROS/repos/resource-provisioning-service/browse/src/main/resources/globaledge_osb.json), so the format of OSB parameters don't bother me, as long as @cetanu got them right.

pkg/orchestration/wiring/edge/types.go
```diff
@@ -0,0 +1,46 @@
...
+	Resource    ResourceParameters  `json:"resource"`
+}
+
+type ResourceParameters struct {
+	Attributes Attributes `json:"attributes"`
+}
+
+type Attributes struct {
+	UpstreamAddress []UpstreamAddress `json:"upstream_address"`
+	UpstreamPort    int32             `json:"upstream_port,omitempty"`
```
ash2k: So the issue here is that the platform user will have parameters with different style for different resource types. As a user, I don't care if it's the platform or its the provider who specifies names. I see that naming is all over the place and I'm not amused by that. FWIW our provider contract page recommends `camelCase` to follow the existing style and Kubernetes style.

@roaanv @wathrog, do we want to do something about it?

ash2k: We have `snake_case`, we have `CapitalCase` and we have `camelCase`.

nilebox: Apart from aesthetics, I don't see this being a problem. The closer resource parameters are to the actual OSB contract, the easier it is to implement an autowiring function, and we probably want to make implementing autowiring functions as simple as possible...

ash2k: "Try to make it pass through" is actually our recommendation, yes. Yes, only aesthetics suffer but that is a significant issue imo.

pkg/orchestration/wiring/edge/types.go
cetanu: nope, it can be 'yes', 'no', or an fqdn

pkg/orchestration/wiring/edge/types.go
cetanu: This field was originally for a single domain, was changed later to allow multiple, bad name

pkg/orchestration/wiring/edge/types.go
nilebox: @cetanu  as discussed with @ash2k in person, the feedback from @roaanv is that we enforce the consistent style in SD, which means that we need to define separate types for SD, and convert to OSB contract internally. I will give it a shot.

ash2k:
You could pre-allocate some slices in the new conversion code and also make the osb-related types private.

nilebox:
All fixed, ready to review.

PR #163, opened 2019-02-13
Refactor to remove legacyconfig object

amckague:
At best legacyconfig belongs to the individual plugins, simplify the
existing wiring contract by removing it from the wiring context.
ec2compute v2 no longer has a v1 version present so remove the extra
package, add an api to allow outside users to set the default set of
tags passed in the entangler context


pkg/orchestration/wiring/ec2compute/v2/plugin.go
```diff
@@ -159,7 +194,27 @@ func constructComputeParameters(origSpec *runtime.RawExtension, iamRoleRef, iamI
 	return rawExtension, false, false, nil
 }
 
-func WireUp(stateResource *orch_v1.StateResource, context *wiringplugin.WiringContext) wiringplugin.WiringResult {
+func New() *WiringPlugin {
+	return &WiringPlugin{
+		DeveloperRole: func(_ voyager.Location) []string {
+			return []string{"arn:aws:iam::123456789012:role/micros-server-iam-MicrosServer-ABC"} //example
```
amckague: Legacy config is a _major_ issue for manual or experimental testing of autowiring functions. You have to use the internal build to access this data. I can probably move most or all of these to envvar lookups. (the way shapes hold things together also requires all the functions be in this repo)

pkg/orchestration/wiring/entangler.go
```diff
@@ -24,14 +23,6 @@ var (
 	serviceInstanceGVK = sc_v1b1.SchemeGroupVersion.WithKind("ServiceInstance")
 )
 
-const (
-	// This is the 'old' Micros name of the environment (e.g. ddev)
-	// We hack this Atlassianism in here rather than putting in user
-	// tags in the configmap because it's at the same level as legacyConfig
-	// (i.e. want to make it obvious they should be removed at the same time).
-	legacyEnvironmentTagName = "environment"
```
amckague: Get to kill some todo's with this move as well

pkg/orchestration/wiring/entangler.go
```diff
@@ -68,12 +59,19 @@ type TagNames struct {
 	EnvironmentTypeTag voyager.Tag
 }
 
+type TagGenerator func(
```
amckague: This API is possible too broad. Please comment on possible alternatives.

pkg/orchestration/wiring/entangler.go
```diff
@@ -158,39 +156,17 @@ func (en *Entangler) Entangle(state *orch_v1.State, context *EntangleContext) En
...
 	for tag, val := range context.ServiceProperties.UserTags {
 		tags[tag] = val
 	}
 
-	tags[en.TagNames.ServiceNameTag] = string(context.ServiceName)
-	tags[en.TagNames.BusinessUnitTag] = context.ServiceProperties.BusinessUnit
-	tags[en.TagNames.ResourceOwnerTag] = context.ServiceProperties.ResourceOwner
-	tags[en.TagNames.EnvironmentTypeTag] = string(location.EnvType)
-	tags[en.TagNames.PlatformTag] = "voyager"
-	tags[legacyEnvironmentTagName] = legacyConfig.MicrosEnv
```
amckague: This lookup is what requires the tags generation to be a function and not just a static map. The other keys could be seen as opinionated defaults.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -98,8 +99,20 @@ func validateScaling(s Scaling) error {
 	return nil
 }
 
+func New() *WiringPlugin {
+	return &WiringPlugin{
+		VPC: func(location voyager.Location) *oap.VPCEnvironment {
+			return oap.ExampleVPC(location.Label, location.Region)
```
amckague: VPCEnvironment has a much smaller surface area than legacy config.

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -158,7 +163,7 @@ func (p *WiringPlugin) WireUp(resource *orch_v1.StateResource, context *wiringpl
 
 // instanceParameters constructs ServiceInstance parameters.
 // sharedDbDep may be nil
-func instanceParameters(resource *orch_v1.StateResource, context *wiringplugin.WiringContext, sharedDbDep *wiringplugin.WiredDependency) ([]byte, bool /* externalErr */, bool /* retriable */, error) {
+func (p *WiringPlugin) instanceParameters(resource *orch_v1.StateResource, context *wiringplugin.WiringContext, sharedDbDep *wiringplugin.WiredDependency) ([]byte, bool /* externalErr */, bool /* retriable */, error) {
```
amckague: @nilebox I opted for a receiver param over an additional function param to avoid clashing with your plans in this place. I would prefer a function param if it doesn't get in the way.

pkg/orchestration/wiring/wiringplugin/types.go
```diff
@@ -122,12 +121,6 @@ type StateContext struct {
 	// from the EntanglerContext.
 	Location voyager.Location
 
-	// LegacyConfig is read by a function specified in the entangler struct.
-	// TODO this is a temporary container for 'stuff that's in Micros config.js'.
```
amckague: gone

pkg/orchestration/wiring/ec2compute/v2/plugin.go
```diff
@@ -2,34 +2,59 @@ package v2
...
+	inputParameterEnvVarName           = "secretEnvVars"
+
+	MaximumServiceNameLength = 26
 )
 
 // HACK: Some tags the EC2 provider doesn't like, because it wants to
 // set them itself... (NB handles business_unit/resource_owner separately)
 // We only really worry about the tags that we're likely to set here
 // (it's ok if the user errors out from the provider).
 var forbiddenTags = map[string]struct{}{
```
ash2k: set of strings?

ash2k:
Legacy configuration should be moved to providers, we have tickets to do that. But I'm not opposed to refactoring this if it makes our life easier.

pkg/orchestration/wiring/postgres/plugin.go
```diff
@@ -54,10 +54,15 @@ type LocationSpec struct {
...
 
 type WiringPlugin struct {
+	Environment func(location voyager.Location) string
 }
 
 func New() *WiringPlugin {
-	return &WiringPlugin{}
+	return &WiringPlugin{
+		Environment: func(_ voyager.Location) string {
+			return "microstestenv"
```
nilebox: shouldn't you accept environment as parameter in constructor instead?

pkg/orchestration/wiring/postgres/plugin.go
nilebox: that's fine

cmd/orchestration/app/controller_constructor.go
```diff
@@ -30,9 +29,9 @@ import (
 )
 
 type ControllerConstructor struct {
-	FlagConfigFile      string
-	GetLegacyConfigFunc func(voyager.Location) *legacy.Config
```
nilebox: The real problem, I think, is that autowiring functions at the moment are not configurable (they are all hardcoded), and the only way to pass configuration is via context and orchestration controller constructor

cmd/orchestration/app/controller_constructor.go
```diff
@@ -88,14 +87,7 @@ func (cc *ControllerConstructor) New(config *ctrl.Config, cctx *ctrl.Context) (*
 		Plugins:         cc.Plugins,
 		ClusterLocation: opts.Location.ClusterLocation(),
 		ClusterConfig:   toClusterConfig(opts.Cluster),
-		TagNames: wiring.TagNames{
```
nilebox: Why this change? Seems unrelated to LegacyConfig. Could you split into multiple PRs please?

pkg/creator/handler.go
```diff
@@ -13,7 +13,7 @@ import (
 	creator_v1 "github.com/atlassian/voyager/pkg/apis/creator/v1"
 	"github.com/atlassian/voyager/pkg/creator/luigi"
 	"github.com/atlassian/voyager/pkg/creator/ssam"
-	ec2compute_common "github.com/atlassian/voyager/pkg/orchestration/wiring/ec2compute/common"
```
nilebox: This is also unrelated to LegacyConfig? separate PR please.

pkg/orchestration/wiring/aws/resources.go
```diff
@@ -34,11 +34,16 @@ const (
 
 // All osb-aws-provider resources are 'almost' the same, differing only in the service/plan names,
 // what they need passed in the ServiceEnvironment.
-var ResourceTypes = map[voyager.ResourceType]wiringplugin.WiringPlugin{
-	DynamoDB: wiringutil.StatusAdapter(Resource(DynamoDB, DynamoDBName, DynamoDBClass, DynamoDBPlan, dynamoDbServiceEnvironment, dynamoDbShapes).WireUp),
-	S3:       wiringutil.StatusAdapter(Resource(S3, S3Name, S3Class, S3Plan, s3ServiceEnvironment, s3Shapes).WireUp),
-	Cfn:      wiringutil.StatusAdapter(Resource(Cfn, CfnName, CfnClass, CfnPlan, CfnServiceEnvironment, cfnShapes).WireUp),
-}
+var (
+	emptyVPC = func(location voyager.Location) *oap.VPCEnvironment {
```
nilebox: How do you construct non-empty VPC and make it environment-specific without LegacyConfig?

pkg/orchestration/wiring/ec2compute/v2/plugin.go
```diff
@@ -159,7 +194,27 @@ func constructComputeParameters(origSpec *runtime.RawExtension, iamRoleRef, iamI
 	return rawExtension, false, false, nil
 }
 
-func WireUp(stateResource *orch_v1.StateResource, context *wiringplugin.WiringContext) wiringplugin.WiringResult {
+func New() *WiringPlugin {
```
amckague: The members of this struct and it's use are the only changes to this class, everything else is just moving code from the extra package.

pkg/orchestration/wiring/ec2compute/v2/plugin.go
nilebox: I don't disagree, it would be nice to replace LegacyConfig with resource-specific configuration.
The problem is, as I said above, that at the moment there is no way to pass configuration to autowiring plugins, and at the moment we are abusing WiringContext and LegacyConfig as a global per-environment configuration.

ash2k: There is a ticket to make autowiring functions accept configuration into their constructor. Even with legacy config completely gone we'd still need this functionality.

nilebox: That's exactly what I meant, if it wasn't clear :)

pkg/orchestration/wiring/aws/resources.go
amckague: We're passing a function type based on public types. I won't be using know authowire plugins in the close repo.

pkg/orchestration/wiring/ec2compute/v2/plugin.go
amckague: We build the list of plugins. All of their "constructors" are accessible. There is not requirement to use known plugins object.

pkg/orchestration/wiring/aws/resources.go
nilebox: I'm sorry, I don't understand this English :(

pkg/orchestration/wiring/ec2compute/v2/plugin.go
nilebox: Then the first trivial change would be to pass LegacyConfig to plugin constructors which use information from it, and remove it from the context? This would be a global LegacyConfig I suppose (not location-specific), but that's ok. It could also be a function providing a LegacyConfig for a given location. This should be a relatively simple isolated change, easy to review.

nilebox: Actually, we already have this function: https://github.com/atlassian/voyager/blob/master/pkg/orchestration/wiring/entangler.go#L161-L173
So we just need to pass this function to plugin constructors, instead of Entangler invoking it and adding to the context. The only problem again is that list of plugins is static at the moment IIRC, i.e. you can't pass a different configuration to it for testing vs production.

amckague: Look at the main function for orchestration. It takes registry.KnownWiringPlugins as an arg of type map[voyager.ResourceType]wiringplugin.WiringPlugin. There is no reason this needs to be the same map inside the close source repo.

pkg/orchestration/wiring/aws/resources.go
amckague: It's not. Follow the code in the orchestration main in _this_ PR.

pkg/orchestration/wiring/postgres/plugin.go
amckague: For all of them yes.

pkg/orchestration/wiring/aws/resources.go
nilebox: I'm confused. If you want to construct plugins dynamically, these static plugin variables simply should not exist, at least in the non-test package.

amckague:
@nilebox @ash2k separate commits for moving common vs removing legacy.

pkg/orchestration/wiring/ec2compute/v2/plugin.go
amckague: Also notice that master of this repo currently stubs legacy config with struct that will nil pointer panic for anything that attempts to use it.

pkg/orchestration/wiring/aws/resources.go
ash2k: I agree with Nail's point. Would be better to move the not-really-ready-to-use instances of plugins (and other "stubby" stuff) into some testing package

pkg/orchestration/wiring/wiringutil/oap/types.go
```diff
@@ -39,6 +38,22 @@ type VPCEnvironment struct {
 	EMRSubnet             string         `json:"emrSubnet,omitempty"`
 }
 
+var ExampleVPC = func(label voyager.Label, region voyager.Region) *VPCEnvironment {
```
ash2k: Should be in some testing package too.

pkg/orchestration/wiring/aws/resources.go
amckague: Think the comment is now outdated.

pkg/orchestration/wiring/wiringutil/oap/types.go
amckague: done

cmd/orchestration/app/main.go
```diff
@@ -23,18 +25,23 @@ const (
 )
 
 func Main() {
-	CustomMain(emptyLegacyConfigFunc, registry.KnownWiringPlugins)
+	CustomMain(registry.KnownWiringPlugins(
```
nilebox: This looks better 👍

nilebox:
@amckague the code looks better now (there is still a lot unrelated changes, so I would prefer to have separate PRs rather than commits, but that's up to you).

My concern though is that if we merge this change, we won't be able to use the latest master until we copy bits from deleted LegacyConfig to a Stash repo and construct all the things there. As alternative, we could keep such code inside this repo for now as a default behavior, and once we rewrite the code in Stash to explicitly pass configuration, we can get rid of this backward compatibility.

i.e. I propose to keep `KnownWiringPlugins` constructless and mark it as deprecated, and add a new method that will accept configuration that we need to migrate to.

cmd/orchestration/app/controller_constructor.go
ychen-atlassian: What about moving the plugins out of wiring/registry? We can pass them as parameters into the controller constructor.

pkg/orchestration/wiring/registry/registry.go
```diff
@@ -16,19 +16,31 @@ import (
...
-	ups.ResourceType:            wiringutil.StatusAdapter(ups.New().WireUp),
-	aws.Cfn:                     aws.ResourceTypes[aws.Cfn],
-	aws.DynamoDB:                aws.ResourceTypes[aws.DynamoDB],
-	aws.S3:                      aws.ResourceTypes[aws.S3],
-	postgres.ResourceType:       wiringutil.StatusAdapter(postgres.New().WireUp),
-	rds.ResourceType:            wiringutil.StatusAdapter(rds.New().WireUp),
-	sqs.ResourceType:            wiringutil.StatusAdapter(sqs.WireUp),
-	asapkey.ResourceType:        wiringutil.StatusAdapter(asapkey.New().WireUp),
-	apiplatformdns.ResourceType: wiringutil.StatusAdapter(platformdns.New().WireUp),
+func KnownWiringPlugins(
```
ychen-atlassian: Sort of related to Nail's comment up https://github.com/atlassian/voyager/pull/163/files#r256667823, we only need to do this because we need to pass a bunch of parameters through the orchestration constructor. When we end up extracting these I assume we'll have a custom orchestration cmd in stash voyager that combines the wiring (like how we combine smith plugins and smith)

pkg/orchestration/wiring/ec2compute/v2/plugin.go
```diff
@@ -2,34 +2,59 @@ package v2
```
ychen-atlassian: Now we can get rid of v2 altogether and just call it ec2compute!

nilebox: just wait for V3 :)
just kidding - yeah, version is not needed anymore (but again, I would prefer this change to be a separate PR)

cmd/orchestration/app/controller_constructor.go
amckague: See the comment about introducing the tag generator. If the tags come from a function there's no need to make their keys come in via config.

pkg/orchestration/wiring/ec2compute/v2/plugin.go
amckague: In our glorious future of go modules that would be breaking the version number in the package path requirement. Also it's separate commits.

pkg/orchestration/wiring/registry/registry.go
ash2k: I have some ideas of how we can have configuration passed to plugins from a shared config file but in a typed way. I view passing functions is a temporary shortcut - instantiating code (orchstration) should not have any knowledge about what configuration is needed i.e. that "configuration" cannot be a function. It should be just a blob of data. But this is one of the next steps.

ash2k: And same with smith - we need a way to pass configuration to plugins https://github.com/atlassian/smith/issues/319 without recompiling smith to change that configuration. We obviously don't want to have a binary per environment, for example.

pkg/orchestration/wiring/wiringutil/oap/wiring.go
```diff
@@ -4,31 +4,17 @@ import (
...
-func MakeServiceEnvironmentFromContext(context *wiringplugin.WiringContext) *ServiceEnvironment {
-	config := context.StateContext.LegacyConfig
-	location := context.StateContext.Location
+func MakeServiceEnvironmentFromContext(context *wiringplugin.WiringContext, vpc *VPCEnvironment) *ServiceEnvironment {
 	serviceProperties := context.StateContext.ServiceProperties
 	alarmEndpoints := PagerdutyAlarmEndpoints(
 		serviceProperties.Notifications.PagerdutyEndpoint.CloudWatch,
 		serviceProperties.Notifications.LowPriorityPagerdutyEndpoint.CloudWatch)
 
 	return &ServiceEnvironment{
```
amckague: Look here for the vpc info

PR #157, opened 2019-02-13
VYGR-461: Migrate RDS to the new OSB wiring helper

nilebox:

PR #164, opened 2019-02-13
Implement saml provider

nickhiggs:
Autowiring function for the saml provider which has parameters name, description and url


pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,97 @@
...
+type autowiringOnlySpec struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	URL         string `json:"url"`
+}
+type finalSpec struct {
+	autowiringOnlySpec
+}
+type WiringPlugin struct {
+	svccatentangler.SvcCatEntangler
```
nilebox: We are migrating from SvcCatEntangler to new helper functions, see #157, #158, #159, #160

pkg/orchestration/wiring/testdata/saml.basic.bundle.yaml
```diff
@@ -0,0 +1,261 @@
...
+      object:
+        apiVersion: servicecatalog.k8s.io/v1beta1
+        kind: ServiceInstance
+        metadata:
+          name: saml
+        spec:
+          clusterServiceClassExternalID: 277d87d1-4cf6-4d96-8bd6-3affb551f21c
+          clusterServicePlanExternalID: 55beb5f4-16f0-4fe5-abcb-95e077271bde
+          parameters:
+            name: test-servicename--testaccount.testenv.testregion--saml
```
nilebox: could you add a test with label too?

pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,137 @@
...
+		bindableEnvVarShape,
+	}, false, false, nil
+}
+
+func instanceParameters(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) ([]byte, bool /* externalErr */, bool /* retriable */, error) {
+	if len(context.Dependencies) > 0 {
+		return nil, false, false, errors.New("saml resource should not have any dependencies")
+	}
+	if resource.Spec == nil {
+		// Don't allow user to set anything they shouldn't
```
ash2k: This is misleading

pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, false, false, errors.New("saml resource is missing user parameters")
+	}
+
+	var spec finalSpec
+
+	err := json.Unmarshal(resource.Spec.Raw, &spec)
+	if err != nil {
+		return nil, false, false, errors.WithStack(err)
```
ash2k: If we cannot unmarshal a blob coming from the user it's an external non-retriable error

pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,137 @@
...
+	// Set default name service--location--resource
+	if spec.Name == "" {
+		label := ""
+		if context.StateContext.Location.Label != "" {
+			label = fmt.Sprintf(".%s", context.StateContext.Location.Label)
+		}
+		spec.Name = fmt.Sprintf("%s--%s.%s.%s%s--%s",
+			context.StateContext.ServiceName,
+			context.StateContext.Location.Account,
+			context.StateContext.Location.EnvType,
```
ash2k: Swap envtype and account?

pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if resource.Spec == nil {
+		return nil, true, false, errors.New("saml resource is missing user parameters")
+	}
+
+	var spec finalSpec
+
+	err := json.Unmarshal(resource.Spec.Raw, &spec)
+	if err != nil {
+		return nil, true, false, errors.WithStack(err)
```
ychen-atlassian: This is probably an internal error because json.Unmarshal shouldn't fail unless the resource spec is wrong, and we already check that via the CRD schema

pkg/trebuchet/server/apiserver/apiserver.go
```diff
@@ -3,7 +3,7 @@ package apiserver
 import (
 	apis_trebuchet "github.com/atlassian/voyager/pkg/apis/trebuchet"
 	"github.com/atlassian/voyager/pkg/apis/trebuchet/install"
-	"github.com/atlassian/voyager/pkg/apis/trebuchet/v1"
+	v1 "github.com/atlassian/voyager/pkg/apis/trebuchet/v1"
```
ychen-atlassian: ?

pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,136 @@
...
+	autowiringOnlySpec
+}
+type WiringPlugin struct {
+}
+
+func New() *WiringPlugin {
+	return &WiringPlugin{}
+}
+
+func (p *WiringPlugin) WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) wiringplugin.WiringResult {
```
ychen-atlassian: If we don't need anything in the WiringPlugin I believe we can just expose the `WireUp` function, see the SQS resource.

pkg/orchestration/wiring/registry/registry.go
```diff
@@ -30,5 +31,6 @@ var KnownWiringPlugins = map[voyager.ResourceType]wiringplugin.WiringPlugin{
 	rds.ResourceType:            wiringutil.StatusAdapter(rds.New().WireUp),
 	sqs.ResourceType:            wiringutil.StatusAdapter(sqs.WireUp),
 	asapkey.ResourceType:        wiringutil.StatusAdapter(asapkey.New().WireUp),
+	saml.ResourceType:           wiringutil.StatusAdapter(saml.New().WireUp),
```
ychen-atlassian: This could probably just be `wiringutil.StatusAdapter(saml.WireUp)` if you remove the struct

pkg/orchestration/wiring/saml/plugin.go
ash2k: Yes, you are right.

pkg/orchestration/wiring/saml/plugin.go
ash2k: I think we'd want to make plugins configurable i.e. to have constructor functions regardless of whether they need any parameters or not. So I think it's fine both ways as we'll likely need to refactor this anyway.

pkg/orchestration/wiring/saml/plugin.go
ychen-atlassian: Oops. Sorry I didn't realize you commented on the same thing.

pkg/trebuchet/server/apiserver/apiserver.go
amckague: Something something go imports is starting to add these. cc @ash2k

ash2k: Yes, new version which we have to use because rules_go needs it. goimports and other things which rules_go needs are all in x/tools repo.

pkg/orchestration/wiring/saml/plugin.go
nickhiggs: You two commented the entirely opposite things 😛

ash2k: I think Yiyang is right, this should be an internal error.

pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return &wiringplugin.WiringResultSuccess{
+		Contract: wiringplugin.ResourceContract{
+			Shapes: shapes,
+		},
+		Resources: []smith_v1.Resource{smithResource},
+	}
+}
+
+func instanceShapes(smithResource *smith_v1.Resource) ([]wiringplugin.Shape, bool, bool, error) {
```
ash2k: Don't need to extract it, just inline into the final literal. No error handling, reads much better.

scottgreenup:
Could you add the ticket number to the title `TNUM: ...`

ash2k:
@ychen-atlassian hey, could you have another look to unblock this PR?

pkg/orchestration/wiring/saml/plugin.go
```diff
@@ -0,0 +1,122 @@
...
+			context.StateContext.ServiceName,
+			context.StateContext.Location.EnvType,
+			context.StateContext.Location.Account,
+			context.StateContext.Location.Region,
+			label,
+			resource.Name,
+		)
+	}
+
+	out, err := json.Marshal(&spec)
```
nilebox: with this approach only parameters defined in the struct will be passed through OSB, is this intended? the code for some other OSB brokers, e.g. OAP, allows for passing through "unknown" parameters that are not defined in the struct.

to me this is fine, especially given the recent agreement to use camel case for all parameters, and it seems that SAM provider is using snake casing (based on response env vars naming). just wanted to double check.

pkg/orchestration/wiring/testdata/saml.label.state.yaml
```diff
@@ -0,0 +1,25 @@
...
+    spec:
+      url: 'https://my-service.prod'
+  - name: kubecompute-simple
+    type: KubeCompute
+    dependsOn:
+    - saml
+    spec:
+      containers:
+      - name: microservice
+        image: "docker.example.com/atlassian/micros-analytics:0.1"
```
nilebox: nit: this is a weird example: image has a tag, and there is a separate tag below. the correct format is not to have a tag in the `image` IIRC.

nilebox: but given that this is a SAML test, this is not critical at all.

PR #165, opened 2019-02-13
Modify monitor to allow arbitrary tests

benbarclay:
This change does a few things:

- Allows us to pass in an arbitrary ServiceDescriptor as a string
- No longer deep checks the generated ServiceInstance(s) as we don't know what will be generated
- Moves the checks to using polling everywhere and uses the ExpectedProcessingTime flag as a deadline for both the test + cleanup


pkg/monitor/monitor.go
```diff
@@ -228,7 +188,7 @@ func (m *Monitor) logFailureDetails(ctx context.Context, retErr error) {
 		service = Service(srv)
 	}
 
-	m.Logger.Error("monitor job failure", serviceInstance, serviceDescriptor, service, zap.Error(retErr))
+	m.Logger.Error("monitor job failure", serviceDescriptor, service, zap.Error(retErr))
```
nilebox: Shall we log summary from reporter as well? :philosoraptor:

cmd/monitor/app/app.go
```diff
@@ -18,9 +18,10 @@ import (
 )
 
 type App struct {
-	RestConfig *rest.Config
-	Logger     *zap.Logger
-	Options    Options
+	RestConfig        *rest.Config
+	Logger            *zap.Logger
+	Options           Options
+	ServiceDescriptor string
```
kopper: nit: it would be nicer if this is part of Options and unmarshalled and parsed to from flag

pkg/monitor/monitor.go
```diff
@@ -204,14 +172,6 @@ func (m *Monitor) waitForServiceDescriptorDeletion(ctx context.Context, name str
 }
 
 func (m *Monitor) logFailureDetails(ctx context.Context, retErr error) {
-	var serviceInstance zap.Field
```
kopper: we need to log these details

kopper: it should be good enough to log full SD and LD as LD contains detailed failure reasons for individual resources

benbarclay: Yeah, the problem is we don't know what ServiceInstances exist for a given ServiceDescriptor. By removing this code path we just always log the full SD and any errors associated with it.

kopper: 1. Of course we do know. If reporter is able to figure out SIs we are able to do it too.
2. We don't need to because all details are in LD anyways.

kopper: I mean there might be some obscure cases when Voyager is so broken than LD has fake news in it, but 🤷‍♂️

nilebox: @benbarclay as I pointed above, we could just log reporter output, but it's going to be HUGE. Adding LD is probably good enough actually, as it contains bubbled up errors from ServiceInstances etc (but SD doesn't!)

cmd/monitor/app/app.go
ash2k: I second that. And `App` should not be parsing it from string, it should accept a typed object.

pkg/monitor/monitor.go
```diff
@@ -75,67 +74,36 @@ func (m *Monitor) Run(ctx context.Context) (retErr error) {
...
-		return errors.Errorf("version parameter missing from ServiceInstance %q in %q", resourceName, namespace)
-	}
-	if foundVersion != expectedVersion {
-		return errors.Errorf("found ServiceInstance with embedded version %q but expecting %q", foundVersion, expectedVersion)
-	}
+		_, ready := cond_v1.FindCondition(sd.Status.Conditions, cond_v1.ConditionReady)
+		if ready == nil {
+			// Probably fine, this can trigger before the SD is processed by the cluster
+			return false, nil
+		}
```
ash2k: generation and observed generation should be checked here in the future.

nilebox: @ash2k SD status still doesn't have `ObservedGeneration` AFAICT :(

pkg/monitor/monitor.go
benbarclay: I'll add LD for now, but that just seems like a hack around the problems with error propagation to SD

pkg/monitor/monitor.go
ash2k: yes, it does not. That is why I'm saying "in the future" :)

pkg/monitor/monitor.go
ash2k: It's not a hack, it's by design. See the relevant doc. You were on that meeting btw 😴

amckague: Dumping reporter output is preferable to reaching for other objects.

pkg/monitor/monitor.go
```diff
@@ -228,7 +186,7 @@ func (m *Monitor) logFailureDetails(ctx context.Context, retErr error) {
 		service = Service(srv)
 	}
 
-	m.Logger.Error("monitor job failure", serviceInstance, serviceDescriptor, service, zap.Error(retErr))
+	m.Logger.Error("monitor job failure", serviceDescriptor, service, zap.Error(retErr))
 }
 
 func (m *Monitor) serviceInstance(ctx context.Context, name, namespace string) (*v1beta1.ServiceInstance, error) {
```
ychen-atlassian: Remove this method if it's no longer being used

pkg/monitor/monitor.go
benbarclay: According to the spec, the SD status is supposed to have the status of each location. If we're dropping the correct status of a location such that we need to pull the LD to get that information, then it's an issue and this is a hack.

pkg/monitor/monitor.go
benbarclay: done

pkg/monitor/monitor.go
nilebox: I tend to agree with Ben. Even if at first we will only show a cluster-local status (i.e. it won't include status of LD from the secondary cluster), it will be a huge improvement already.

kopper: SD contains status of each location, but no resource statuses. Thus it is useless.

cmd/monitor/app/app.go
```diff
@@ -18,9 +19,10 @@ import (
...
```
kopper: not used anymore?

benbarclay: Good catch

kopper:
This PR when merged will break deployments? (because there is no config yet)

benbarclay:
I plan to merge this into the upstream repo with the config files at the same time

cmd/monitor/app/options.go
```diff
@@ -4,17 +4,19 @@ import (
...
 	"sigs.k8s.io/yaml"
 )
 
 type Options struct {
-	ServiceDescriptorName  string                 `json:"name"`
-	Location               options.Location       `json:"location"`
-	ExpectedProcessingTime time.Duration          `json:"expected_processing_time"`
-	ServiceSpec            creator_v1.ServiceSpec `json:"service_spec"`
+	ServiceDescriptorName  string                     `json:"name"`
+	ServiceDescriptor      *comp_v1.ServiceDescriptor `json:"service_descriptor"`
```
ash2k: Ideally we should use camel case everywhere but I see that is already not the case :( Would be great to change that in a separate PR.

pkg/monitor/monitor.go
```diff
@@ -62,80 +67,43 @@ func (m *Monitor) Run(ctx context.Context) (retErr error) {
...
 	}
 
 	m.Logger.Info("Initialized ServiceDescriptor for synthetic checks")
 
-	// Wait until the corresponding ServiceInstance is created.
-	// Sleeps for a short period of time, which for now should be enough I think
-	time.Sleep(m.ExpectedProcessingTime)
-
-	err = m.verifyUpsServiceInstance(ctx, m.ServiceDescriptorName, m.Version)
+	err = wait.PollImmediate(pollDelay, m.ExpectedProcessingTime, m.verifyServiceDescriptorStatus(m.ServiceDescriptorName))
```
ash2k: We can avoid polling to get immediate results using `UntilWithSync()` from `"k8s.io/client-go/tools/watch"`. See how it is used in Smith's integration tests https://github.com/atlassian/smith/blob/e367e27c0d09db8c88715fd76122be086bacd15a/it/utils_for_tests.go#L89-L99

PR #167, opened 2019-02-13
NONE: 1/? Fix up golangci-lint errors

scottgreenup:
**What?**
I've implemented https://github.com/golangci/golangci-lint#configuration as a Bazel rule and run some of the linters and fixed their issues in this PR.
The Bazel rule was local, and is pending review: https://github.com/atlassian/bazel-tools/pull/48

**Why?**
Some of these issues are interesting and are worth fixing. Others are just to make the linter happy and keep our code consistent. We want golangci-lint as it is proving to be a lot faster


pkg/orchestration/wiring/wiringutil/knownshapes/bindable_iam_accessible.go
```diff
@@ -26,7 +26,7 @@ type BindableIamAccessibleData struct {
 	IAMPolicySnippet libshapes.BindingSecretProtoReference
 }
 
-func NewBindableIamAccessible(resourceName smith_v1.ResourceName, IAMPolicySnippetPath string) *BindableIamAccessible {
+func NewBindableIamAccessible(resourceName smith_v1.ResourceName, iamPOlicySnippetPath string) *BindableIamAccessible {
```
ash2k: typo

scottgreenup:
DO NOT MERGE - We should coordinate all repos to land around the same time. To avoid linting errors causing problems up the chain. If a dependant isn't using gometalinter, but the top level is using the linter on it, then it could fail.

PR #168, opened 2019-02-14
VYGR-461: Delete SvcCatEntangler

nilebox:
all pull requests have been migrated to new OSB helper functions, it's safe to delete `SvcCatEntangler`.


ychen-atlassian:
While I like the removal of a bunch of code, I didn't get a chance to look at the other PRs and it would be nice to deduplicate our code inside our wiring plugins which boils down to:

1. Construct service instance
2. Call `instanceParameters` to get the spec
3. Construct the smith resource and shapes
4. Return the resource contract containing the instance and shapes.

nilebox:
> it would be nice to deduplicate our code inside our wiring plugins which boils down to

@ychen-atlassian the problem is that with SvcCatEntangler it was hard to see the code flow, because it looked like:
1. Do something
2. Invoke a callback for instance creation
3. Do something
4. Invoke a callback for references
5. Do something
6. Invoke a callback for shapes
...

If you never implemented an OSB autowiring plugin before, it's very hard to understand the purpose of 5 callback functions that you have to define in the plugin constructor. So with new helper functions, the logic is opposite: the interface is implemented in the specific plugin, which invokes some helper functions to reduce boilerplate. If we can come up with more helper function to reduce new boilerplate, I'm all for it.

ychen-atlassian:
I like that we got rid of a lot of the complexity. Maybe SvcCatEntangler was just leaky and evolved a bit messy? We could have called it something like `ServiceInstanceWiringCreator`

With your changes all that is needed is

- A function that generates the spec of your ServiceInstance
- A function that returns the shapes.
- No embeddeding of structs

Something like

```
func (p *WiringPlugin) WireUp(resource *orch_v1.StateResource, context *wiringplugin.WiringContext) wiringplugin.WiringResult {
   wiring := wiring_util.SimpleServiceInstanceWiring(resource, clusterServiceClassExternalID, clusterServicePlanExternalID)


   return wiring.CreateWiring(instanceParameters, instanceShapes)

}
```

PR #173, opened 2019-02-14
Auto-bump dependencies

atlassian-bamboo-agent:


ash2k:
Fix code to compile with new service catalog.

PR #174, opened 2019-02-14
Deduplicate code for simple wiring

nilebox:
As @ychen-atlassian suggested, extracted repetitive code for simple wiring.

### DIFF
--- a/pkg/orchestration/wiring/wiringutil/util.go
+++ b/pkg/orchestration/wiring/wiringutil/util.go
@@ -1,7 +1,6 @@
 package wiringutil
 
 import (
-	"k8s.io/apimachinery/pkg/runtime"
 	"reflect"
 	"strings"
 
@@ -9,6 +8,7 @@ import (
 	"github.com/atlassian/voyager/pkg/orchestration/wiring/wiringplugin"
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // Merge two maps, only use loser's fields if those fields are missing in winner
--- a/pkg/orchestration/wiring/wiringutil/util.go
+++ b/pkg/orchestration/wiring/wiringutil/util.go
@@ -1,9 +1,12 @@
 package wiringutil
 
 import (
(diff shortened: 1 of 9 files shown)


pkg/orchestration/wiring/wiringutil/util.go
```diff
@@ -43,3 +46,21 @@ func StripJSONFields(obj map[string]interface{}, badStruct interface{}) {
 		}
 	}
 }
+
+// Return wiring result with a single object
+func SingleWiringResult(name smith_v1.ResourceName, obj runtime.Object, shapes []wiringplugin.Shape, references []smith_v1.Reference) *wiringplugin.WiringResultSuccess {
```
ash2k: I don't think it's super valuable to replace a struct literal with a function. Yes, it's "less code" but it adds indirection without incapsulating any logic. I'm happy for this to be merged, I'm just saying it's not a clear win.

nilebox: agreed, I'm on the fence myself about this change.
@ychen-atlassian could you please review and check if that's what you had in your mind?

ychen-atlassian: I don't think this change is super valuable either. I was thinking more about pulling out almost the *whole* WireUp function, which includes all the bits to do with service instance too.

I.e. svccatentangler v2, except cleaner. Our WireUp function should be using a util that makes sense. In pseudocode probably something that looks like:

```
WireUp() {
   myInstanceSpec := instanceSpec()
   myShapes := instanceShapes()

   return wiring_util.ServiceInstanceWiringResult(myInstanceSpec, myShapes)
}
```

The value here is deduplicating the whole function. I'm not too fussed about it. There's not much value - imagine making the shapes change without svccatentangler - we would have needed to make the same change to every resource, but we needed to do this anyway to pass the shapes through, so it's probably fine.

ychen-atlassian:
I'm okay with leaving the code as is. I literally just was thinking about having svccatentangler except as a util method rather than as a struct that you embed to get WireUp (which imo was the confusing part)

PR #177, opened 2019-02-18
NONE: Remove unused function

ash2k:


ash2k:
Build has failed because of a flaky test. I have found the cause, will send a PR (non-deterministic variable checking because of the map iteration order).

PR #178, opened 2019-02-18
Deterministic map expansion order

ash2k:
Fix map iteration order bug found while looking at the test flakiness for https://github.com/atlassian/voyager/pull/177.

### DIFF
--- a/pkg/util/templating/templating.go
+++ b/pkg/util/templating/templating.go
@@ -2,6 +2,7 @@ package templating
 
 import (
 	"encoding/json"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/util/templating/templating.go
```diff
@@ -243,11 +250,8 @@ func Merge(winner, loser interface{}) (interface{}, error) {
 // use in the applicable (child)map
 func FindInMapRecursive(src map[string]interface{}, keyPath []string) (interface{}, error) {
```
ash2k: Changes below are just cleanup

PR #180, opened 2019-02-18
Add annotation to disable validation of known locations

nilebox:
This will allow us to use a single SD for testing and production environments

### DIFF
--- a/pkg/replication/admission.go
+++ b/pkg/replication/admission.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/replication/admission.go
```diff
@@ -41,7 +42,8 @@ const (
 	updatedKey = "mutationTimestamp"
 	hashKey    = "mutationHash"
 
-	ReplicateKey = "replicate"
+	ReplicateKey              = "replicate"
+	ValidateKnownLocationsKey = "validateKnownLocations"
```
nilebox: For some reason all SD annotations listed above are not fully qualified (do not include `voyager.atl-paas.net` prefix), so I added a new annotation to follow established pattern.

We should probably fix this and migrate to qualified annotations at some point.
/cc @ash2k @benbarclay

PR #182, opened 2019-02-19
Update Ctrl version, propogate external errors up to ctrl layer

ychen-atlassian:
This bumps the version of Ctrl.

As a result, I also needed to change everything to fit the new interface, which meant bubbling external errors up to the ctrl layer instead of hiding them.

### DIFF
--- a/pkg/composition/controller.go
+++ b/pkg/composition/controller.go
@@ -69,7 +69,7 @@ func (c *Controller) Run(ctx context.Context) {
 	<-ctx.Done()
 }
 
-func (c *Controller) Process(ctx *ctrl.ProcessContext) (bool /* retriable */, error) {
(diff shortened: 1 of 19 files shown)


pkg/composition/controller.go
```diff
@@ -111,15 +111,15 @@ func (c *Controller) Process(ctx *ctrl.ProcessContext) (bool /* retriable */, er
...
 		}
 
 		// status update failed (real error)
 		if handleResultErr != nil {
-			return handleResultRetriable, handleResultErr
+			logger.Error("Failed to set Service Descriptor status", zap.Error(handleResultErr))
 		}
 
-		return false, nil
+		return true, handleResultRetriable, err
```
amckague: Should look at returning a list of errors.

ash2k:
Build is failing because of a flaky test that I fixed recently but this branch was created before that.

PR #183, opened 2019-02-19
MICROS-6976 PDB name should change if configuration changes

halcyonCorsair:
PodDisruptionBudget is immutable and can't be modified by smith, therefore if we change the config, we need to give the object a new name.

### DIFF
--- a/pkg/orchestration/wiring/k8scompute/plugin.go
+++ b/pkg/orchestration/wiring/k8scompute/plugin.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"strings"
 
 	smith_v1 "github.com/atlassian/smith/pkg/apis/smith/v1"
 	"github.com/atlassian/voyager"
@@ -334,9 +335,9 @@ func (p *WiringPlugin) WireUp(resource *orch_v1.StateResource, context *wiringpl
 	podSpec := buildPodSpec(containers, serviceAccountNameRef.Ref(), affinity)
 
 	// Add pod disruption budget
-	pdbSpec := buildPodDisruptionBudgetSpec(labelMap, spec.Scaling.MinReplicas)
+	pdbSpec, pdbNamePostfix := buildPodDisruptionBudgetSpec(labelMap, spec.Scaling.MinReplicas)
 	pdb := smith_v1.Resource{
-		Name: wiringutil.ResourceName(resource.Name, pdbPostfix),
+		Name: wiringutil.ResourceName(resource.Name, pdbPostfix, pdbNamePostfix),
 		Spec: smith_v1.ResourceSpec{
 			Object: &policy_v1.PodDisruptionBudget{
 				TypeMeta: meta_v1.TypeMeta{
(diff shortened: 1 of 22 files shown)


pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -334,9 +335,9 @@ func (p *WiringPlugin) WireUp(resource *orch_v1.StateResource, context *wiringpl
 	podSpec := buildPodSpec(containers, serviceAccountNameRef.Ref(), affinity)
 
 	// Add pod disruption budget
-	pdbSpec := buildPodDisruptionBudgetSpec(labelMap, spec.Scaling.MinReplicas)
+	pdbSpec, pdbNamePostfix := buildPodDisruptionBudgetSpec(labelMap, spec.Scaling.MinReplicas)
 	pdb := smith_v1.Resource{
-		Name: wiringutil.ResourceNameWithPostfix(resource.Name, pdbPostfix),
+		Name: wiringutil.ResourceNameWithPostfix(resource.Name+"--"+voyager.ResourceName(pdbNamePostfix), pdbPostfix),
```
halcyonCorsair: I didn't add you as a reviewer, but do you object to this @ash2k ?

ash2k: I think it's the right way to do it, but we can do it in a cleaner way. I'll send a PR with small refactoring of helper functions and then you can use the better version of it.

pkg/orchestration/wiring/k8scompute/plugin.go
```diff
@@ -596,9 +597,10 @@ func buildAntiAffinity(labelMap map[string]string) *core_v1.PodAntiAffinity {
 	}
 }
 
-func buildPodDisruptionBudgetSpec(labelMap map[string]string, minReplicas int32) policy_v1.PodDisruptionBudgetSpec {
+func buildPodDisruptionBudgetSpec(labelMap map[string]string, minReplicas int32) (policy_v1.PodDisruptionBudgetSpec, string) {
 	// Make sure we can drain nodes if the minReplicas is < 3
 	var minAvailable string
+	var postfix string
```
Burr: nit: no need to declare postfix here, can be declared below when you set the value

pkg/orchestration/wiring/k8scompute/plugin.go
ash2k: https://github.com/atlassian/voyager/pull/184

ash2k: After this is merged, you'll be able to do this: `ResourceName(resource.Name, pdbPostfix, pdbNamePostfix)`

ash2k: merged

ychen-atlassian (dismissed):

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Fixed.

pkg/orchestration/wiring/k8scompute/plugin.go
halcyonCorsair: Fixed.

Burr (approved):

ash2k (approved):

halcyonCorsair:
Note for future self: In playground I tested adding a second pod disruption budget with the same label selectors, and discussed it with KITT, if a new one exists for a short period of time before the old one is deleted by smith, we there won't be any major issues.

PR #185, opened 2019-02-20
Fix error message when deleting existing container

mmatusiak-atlassian:
For ssam containers that were somehow created without obeying the naming scheme we currently block deleting them. This allows cleaning them up.


pkg/creator/ssam/client.go
```diff
@@ -326,8 +326,8 @@ func (c clientImpl) DeleteContainer(ctx context.Context, containerShortName stri
 	logger.Debug("SSAM GetContainer", zappers.ContainerShortName(containerShortName))
 
 	if err := validateShortName(containerShortName); err != nil {
-		return err
-
+		logger.Warn("SSAM container name %q does not obey naming scheme, will attempt to delete anyway",
```
nilebox: you should log the error as well

mmatusiak-atlassian:
I'm retracting this because the ssam container doesn't actually exist so this solves nothing. My bad for not checking that.

PR #186, opened 2019-02-21
Only keep existing mutation timestamp when user is replication

benbarclay:


pkg/replication/admission.go
```diff
@@ -37,6 +38,9 @@ import (
 )
 
 const (
+	// User account this component uses
```
nilebox: "this component"? replication admission is a separate component from replication controller...
I would rephrase comment to explicitly mention replication controller.

pkg/replication/admission.go
```diff
@@ -332,8 +336,12 @@ func (ac *AdmissionContext) validateLocationsAndTransforms(sd *comp_v1.ServiceDe
 	return RejectionMessage(strings.Join(rejectionMessages, ", "))
 }
 
-func createAnnotationPatch(operation admissionv1beta1.Operation, newSD *comp_v1.ServiceDescriptor) ([]byte, error) {
-	ts, err := generateUpdated(operation, newSD)
+func isReplication(userInfo authenticationv1.UserInfo) bool {
```
nilebox: `isReplicationController`?

pkg/replication/admission.go
```diff
@@ -489,15 +497,14 @@ func addAnnotation(key, value string) util.Patch {
 	}
 }
 
-func generateUpdated(operation admissionv1beta1.Operation, newSD *comp_v1.ServiceDescriptor) (time.Time, error) {
-	if operation != admissionv1beta1.Update {
+func generateUpdated(isReplication bool, operation admissionv1beta1.Operation, newSD *comp_v1.ServiceDescriptor) (time.Time, error) {
+	if operation != admissionv1beta1.Update || !isReplication {
```
nilebox: Hmm... So this code is written in a way that if replication controller doesn't set the timestamp (why? is that possible?), we will generate a timestamp.

Maybe instead we could always parseTimestamp for replication controller and return error if annotation is missing?

pkg/replication/admission.go
benbarclay: I view replication as an entire subsystem. The implementation is split across a controller and a couple of webhooks. I take it you see it differently?

pkg/replication/admission.go
benbarclay: afaict the only situation we *don't* want to generate a timestamp is if the replication controller is doing an update on an existing thing. In all other cases we do want to generate a timestamp. Isn't that what's going on here?

nilebox: Yes, but what is the case when replication controller doesn't provide a timestamp? I guess to rephrase the requirement: replication controller _must_ set timestamp and we don't touch it (but validate it's set). In all other cases we generate timestamp.
This way it will be easier to catch bugs in replication controller I think

pkg/replication/admission.go
nilebox: Fair enough, but in this case it's replication controller request to API server being intercepted by a webhook, so it seems useful to be more clear about what we do.

pkg/replication/admission.go
benbarclay: Fair call, I'll fix that up

nilebox: Also not sure if we actually still need to check for `operation != admissionv1beta1.Update`? I find this code hard to understand. My idea:
```
if isReplication {
  ts, exists := newSD.Annotations[updatedKey]
  if !exists {
    return nil, errors.New("replication controller must set the mutation timestamp")  
  }
  return parseTimestamp(ts)
}

return time.Now(), nil
```

pkg/replication/admission.go
```diff
@@ -489,16 +497,15 @@ func addAnnotation(key, value string) util.Patch {
...
+	if isReplication {
+		ts, exists := newSD.Annotations[updatedKey]
+		if !exists {
+			return time.Now(), errors.New("replication controller must set the mutation timestamp")
+		}
+		return parseTimestamp(ts)
 	}
-
-	return parseTimestamp(ts)
+	return time.Now(), nil
```
nilebox: Sorry, I think we may need to keep the timestamp if hash hasn't changed, e.g. when someone adds/deletes annotation or finalizer.
So you need to pass a flag if hash has changed?

